PR #11, opened 2025-10-05
Default learning_rate_strategy to ConstantLR for simplicity

omrishefi:


neuralhydrology/utils/config.py
```diff
@@ -572,8 +572,8 @@ def lead_time(self) -> int:
         return self._cfg.get("lead_time", 0)
 
     @property
-    def learning_rate_strategy(self) -> str|None:
-        return self._cfg.get("learning_rate_strategy", None)
+    def learning_rate_strategy(self) -> str:
```
amitmarkel, line 575: Now that it's not None, the code that calls this doesn't need to handle None necessarily.

amitmarkel (approved):

neuralhydrology/utils/config.py
omrishefi: If the value given in learning_rate_strategy is invalid, we still need to fail the code.

amitmarkel: To clarify, you misread my comment. Apparently the code checks via else/default and raises. Previously it selected something by default. All I said was to not select something by default (None).

PR #36, opened 2025-10-20
Allow define different activation layers in embedding  config

omrishefi:


googlehydrology/modelzoo/fc.py
```diff
@@ -32,7 +32,7 @@ class FC(nn.Module):
         Number of input features.
     hidden_sizes : list[int]
         Size of the hidden and output layers.
-    activation : str, optional
+    activation : str | list[str], optional
```
amitmarkel, line 35: When can it be a str? Always a list no?

amitmarkel (approved):

googlehydrology/modelzoo/fc.py
omrishefi: This is an internal class, which can be called with str directly

PR #37, opened 2025-10-20
Create FC embedding layer dynamically

omrishefi:


googlehydrology/modelzoo/mean_embedding_forecast_lstm.py
```diff
@@ -129,6 +119,28 @@ def __init__(self, cfg: Config):
 
         self._reset_parameters()
 
+    def _create_fc(self, embedding_spec: dict|None, input_size: int) -> FC:
```
amitmarkel: let's use asserts with msgs

amitmarkel (approved):

PR #40, opened 2025-10-21
Merge calculation of embeddings into a single function

omrishefi:


amitmarkel (commented):
Maybe there's an error with the pr, it looks like only a rename of the function

amitmarkel (approved):

PR #42, opened 2025-10-21
Adjust config for benchmark dataset train.

amitmarkel:


config/multimet_mean_embedding_forecast_lstm.yml
```diff
@@ -25,18 +25,18 @@ clip_targets_to_zero:
...
 metrics:
 - NSE
-- KGE
-- Alpha-NSE
-- Beta-NSE
-- Beta-KGE
-- Peak-Timing
-- Missed-Peaks
+# - KGE
+# - Alpha-NSE
```
omrishefi: Keep these, it doesn't takes time to calculate them.

omrishefi (dismissed):
See my comment

config/multimet_mean_embedding_forecast_lstm.yml
amitmarkel: They're noise on tensorboard. If we want, we can always run evaluate and get these.

omrishefi: But in this point, it's important we have all data, since we're looking for ideas how to improve the NSE. We shouldn't remove them at this point.

amitmarkel: Ok (done) but I disagree here. KGE acts like NSE and all the other metrics at this point aren't relevant to us. I don't see how any of the other metrics helps us here other than making it hard to find graphs on tensorboard.

amitmarkel: Still need approval due to changes

omrishefi (approved):

PR #45, opened 2025-10-21
Filter out warning logs that were already reported.

amitmarkel:
This affects mainly `infer` mode where the same warnings are printed several times on every iteration otherwise.


omrishefi (commented):
There is no need in this change

amitmarkel:
Without the change it's harder to test the infer mode, and we can't address the warnings yet.

amitmarkel:
Added print_warnings_once flag

omrishefi (approved):

PR #48, opened 2025-10-22
Add a small utility function to group features by prefix

omrishefi:


googlehydrology/utils/configutils.py
```diff
@@ -15,7 +15,7 @@
 """Utility script to generate config files from a base config and a defined set of variations"""
 import itertools
 from pathlib import Path
-
+from collections import defaultdict
```
amitmarkel: Don't use default dict. When used afterwards, keys would be added instead of key error.

I worked on removing defauptdict from the library and the original authors even had todos for that. 

Use dict's setdefault

googlehydrology/utils/configutils.py
```diff
@@ -66,3 +66,15 @@ def flatten_feature_list(data: list[str] | list[list[str]] | dict[str, list[str]
...
+
+
+def group_by_prefix(features: list[str]) -> dict[str, list[str]]:
+    """
+    Groups list of features according to the prefix of each feature.
+    Assumes the prefix is written with an underscore at the start of the feature name.
+    """
+    result = defaultdict(list)
+    for feature in features:
+        prefix = feature.split('_')[0]
```
amitmarkel: Partition instead of split

amitmarkel (dismissed):

amitmarkel (approved):

PR #53, opened 2025-10-26
Allow activation to be a single value

omrishefi:


googlehydrology/utils/config.py
```diff
@@ -1017,10 +1017,14 @@ def verbose(self) -> int:
     def _get_embedding_spec(self, embedding_spec: dict | None) -> dict | None:
         if embedding_spec is None:
             return None
+        hiddens = self._as_default_list(embedding_spec.get('hiddens', []))
+        activation = embedding_spec.get('activation', 'tanh')
+        if not isinstance(activation, list):
+            activation = [activation] * len(hiddens)
```
amitmarkel, line 1025: let's assert length eq early on. eg

```
if isinstance list:
  assert len(activation) == len(hiddens)
else:
  .. = [..] * len(hiddens)
```

amitmarkel (dismissed):

amitmarkel (approved):

PR #59, opened 2025-10-29
Add weight_init_opts: first add xavier for ih lstm and fc

amitmarkel:
We should test inits that aren't necessarily uniform random or zeros, such as [xavier_uniform_](https://docs.pytorch.org/docs/stable/nn.init.html#torch.nn.init.xavier_uniform_) for ih (input-to-hidden) weights and [orthogonal_](https://docs.pytorch.org/docs/stable/nn.init.html#torch.nn.init.orthogonal_) for hh (hidden to hidden).

NSE seems starting off a bit better, e.g. 10 min train:
Screenshot 2025-10-29 at 13 55 46

### DIFF
--- a/googlehydrology/modelzoo/fc.py
+++ b/googlehydrology/modelzoo/fc.py
@@ -36,12 +36,14 @@ class FC(nn.Module):
         Activation function for intermediate layers, default tanh.
     dropout : float, optional
         Dropout rate in intermediate layers.
+    fc_xavier : bool, optional
+        Whether to use xavier as init method.
     """
 
-    def __init__(self, input_size: int, hidden_sizes: list[int], activation: str | list[str] = 'tanh', dropout: float = 0.0, advanced: bool = False):
+    def __init__(self, input_size: int, hidden_sizes: list[int], activation: str | list[str] = 'tanh', dropout: float = 0.0, fc_xavier: bool = False):
         super(FC, self).__init__()
 
-        self._advanced = advanced
+        self._xavier = fc_xavier
 
         if len(hidden_sizes) == 0:
             raise ValueError('hidden_sizes must at least have one entry to create a fully-connected net.')
@@ -92,7 +94,7 @@ class FC(nn.Module):
             if isinstance(layer, nn.modules.linear.Linear):
                 n_in = layer.weight.shape[1]
                 gain = np.sqrt(3 / n_in)
-                if self._advanced:
+                if self._xavier:
                     nn.init.xavier_uniform_(layer.weight, gain)
                 else:
                     nn.init.uniform_(layer.weight, -gain, gain)
--- a/googlehydrology/modelzoo/fc.py
+++ b/googlehydrology/modelzoo/fc.py
@@ -38,10 +38,10 @@ class FC(nn.Module):
         Dropout rate in intermediate layers.
     """
 
-    def __init__(self, input_size: int, hidden_sizes: list[int], activation: str | list[str] = 'tanh', dropout: float = 0.0, xavier: bool = False):
+    def __init__(self, input_size: int, hidden_sizes: list[int], activation: str | list[str] = 'tanh', dropout: float = 0.0, advanced: bool = False):
         super(FC, self).__init__()
 
-        self._xavier = xavier
+        self._advanced = advanced
 
         if len(hidden_sizes) == 0:
             raise ValueError('hidden_sizes must at least have one entry to create a fully-connected net.')
@@ -92,7 +92,7 @@ class FC(nn.Module):
             if isinstance(layer, nn.modules.linear.Linear):
                 n_in = layer.weight.shape[1]
                 gain = np.sqrt(3 / n_in)
-                if self._xavier:
+                if self._advanced:
                     nn.init.xavier_uniform_(layer.weight, gain)
                 else:
                     nn.init.uniform_(layer.weight, -gain, gain)
--- a/googlehydrology/modelzoo/fc.py
+++ b/googlehydrology/modelzoo/fc.py
@@ -38,8 +38,10 @@ class FC(nn.Module):
         Dropout rate in intermediate layers.
     """
 
-    def __init__(self, input_size: int, hidden_sizes: list[int], activation: str | list[str] = 'tanh', dropout: float = 0.0):
+    def __init__(self, input_size: int, hidden_sizes: list[int], activation: str | list[str] = 'tanh', dropout: float = 0.0, xavier: bool = False):
         super(FC, self).__init__()
+
+        self._xavier = xavier
 
         if len(hidden_sizes) == 0:
             raise ValueError('hidden_sizes must at least have one entry to create a fully-connected net.')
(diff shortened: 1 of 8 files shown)


googlehydrology/modelzoo/handoff_forecast_lstm.py
```diff
@@ -115,11 +117,16 @@ def __init__(self, cfg: Config):
...
-        if self.cfg.initial_forget_bias is not None:
-            self.hindcast_lstm.bias_hh_l0.data[
-                self.hindcast_hidden_size:2*self.hindcast_hidden_size] = self.cfg.initial_forget_bias
-            self.forecast_lstm.bias_hh_l0.data[
-                self.forecast_hidden_size:2*self.forecast_hidden_size] = self.cfg.initial_forget_bias
+        for name, param in self.hindcast_lstm.named_parameters():
+            if 'weight' in name and self.cfg.use_xavier_init:
+                nn.init.xavier_uniform_(param.data)
+            elif 'bias' in name and self.cfg.initial_forget_bias is not None:
+                nn.init.constant_(param.data, self.cfg.initial_forget_bias)
```
omrishefi: Please split to two different PRs

googlehydrology/modelzoo/handoff_forecast_lstm.py
```diff
@@ -115,11 +117,16 @@ def __init__(self, cfg: Config):
...
         """Special initialization of certain model weights."""
-        if self.cfg.initial_forget_bias is not None:
-            self.hindcast_lstm.bias_hh_l0.data[
-                self.hindcast_hidden_size:2*self.hindcast_hidden_size] = self.cfg.initial_forget_bias
-            self.forecast_lstm.bias_hh_l0.data[
-                self.forecast_hidden_size:2*self.forecast_hidden_size] = self.cfg.initial_forget_bias
+        for name, param in self.hindcast_lstm.named_parameters():
+            if 'weight' in name and self.cfg.use_xavier_init:
+                nn.init.xavier_uniform_(param.data)
+            elif 'bias' in name and self.cfg.initial_forget_bias is not None:
```
omrishefi: Please verify it's the same result exactly

googlehydrology/modelzoo/handoff_forecast_lstm.py
amitmarkel: Done

googlehydrology/modelzoo/handoff_forecast_lstm.py
amitmarkel: The odd hacking into the lstm is actually correct: in hiddensize quantiles it's inputs, forget, cell, outputs. Next PR I'll see if there's a pytorch function that returns the appropriate slice instead.

amitmarkel (commented):
Done

googlehydrology/modelzoo/fc.py
```diff
@@ -38,9 +38,11 @@ class FC(nn.Module):
         Dropout rate in intermediate layers.
     """
 
-    def __init__(self, input_size: int, hidden_sizes: list[int], activation: str | list[str] = 'tanh', dropout: float = 0.0):
+    def __init__(self, input_size: int, hidden_sizes: list[int], activation: str | list[str] = 'tanh', dropout: float = 0.0, advanced: bool = False):
```
omrishefi: advanced isn't a good name since it doesn't explain what it does. Better either pass an enum of the initialization, or call is xavier_init explicitly.

amitmarkel: It's going to use xavier and orthogonal. So both of the options don't fit. Also we can always rename it. It's due to splitting PRs. I don't want to refactor everything multiple times just because we're splitting PRs.

Please understand just at this second this sets xavier, but I intend to set other things right away as well. As I wrote here and in chat.

omrishefi: But that's a bad name. We can't insert it as is.

amitmarkel: Again we can rename it next. Also could you suggest a name? Not just "xavier".

The name just needs to convey that init isn't the simple uniform random init.

Regardless, "Better is better than best" - we better iterate more than standstill on speculations.

omrishefi: Please call is xavier_init

amitmarkel: Again that isn't a suitable name. I meant a name that doesn't include "xavier" since the logic follows next to call it like `weights_init__lstm_xavier__fc_orthogonal` which isn't practical.

However. Here's a compromise so we may continue. I changed the flag to a list of possible strings, and changed the name to something else that's still flexible w.r.t. adding options to it.

omrishefi (dismissed):

googlehydrology/modelzoo/fc.py
omrishefi: Passing an enum is much more readable now. LGTM.

googlehydrology/modelzoo/fc.py
```diff
@@ -36,11 +36,15 @@ class FC(nn.Module):
         Activation function for intermediate layers, default tanh.
     dropout : float, optional
         Dropout rate in intermediate layers.
+    fc_xavier : bool, optional
+        Whether to use xavier as init method.
     """
 
-    def __init__(self, input_size: int, hidden_sizes: list[int], activation: str | list[str] = 'tanh', dropout: float = 0.0):
+    def __init__(self, input_size: int, hidden_sizes: list[int], activation: str | list[str] = 'tanh', dropout: float = 0.0, fc_xavier: bool = False):
```
omrishefi: Maybe call is fc_xavier_init

googlehydrology/modelzoo/handoff_forecast_lstm.py
```diff
@@ -16,13 +16,16 @@
...
+from more_itertools import flatten
 
-from googlehydrology.utils.config import Config
+from googlehydrology.utils.config import Config, WeightInitOpt
 from googlehydrology.modelzoo.basemodel import BaseModel
 from googlehydrology.modelzoo.fc import FC
 from googlehydrology.modelzoo.head import get_head
 from googlehydrology.modelzoo.inputlayer import InputLayer
 
+LSTM_IH_XAVIER = WeightInitOpt.LSTM_IH_XAVIER
```
omrishefi, line 27: It's unnecessary to define these in the class level. You can use the enum directly. UTY.

googlehydrology/modelzoo/fc.py
amitmarkel: Just noting it's a list. An enum wouldn't be sufficiently readable since need all combinations in the enum. For example `LSTM_XAVIER__FC_ORTHOGONAL__<third thing>__...` etc.

googlehydrology/modelzoo/fc.py
amitmarkel: Done

googlehydrology/modelzoo/handoff_forecast_lstm.py
amitmarkel: It makes the code below shorter to read. Also aligns with the library's style: e.g. to import symbols from other modules like `get_head`, `InputLayer` etc, it's the same principle.

We could refactor the library to only import symbols that are modules as style in separate though.

amitmarkel:
Thanks for the comments, addressed them, PTAL

omrishefi (approved):

PR #60, opened 2025-10-29
Generalize the embedding network

omrishefi:
Building the embedding FC networks dynamically based on the provided feature list.

### DIFF
--- a/googlehydrology/modelzoo/mean_embedding_forecast_lstm.py
+++ b/googlehydrology/modelzoo/mean_embedding_forecast_lstm.py
@@ -23,6 +23,7 @@ from googlehydrology.modelzoo.basemodel import BaseModel
 from googlehydrology.modelzoo.fc import FC
 from googlehydrology.modelzoo.head import get_head
 from googlehydrology.utils.config import Config
+from googlehydrology.utils.configutils import group_by_prefix
 
 
 class MeanEmbeddingForecastLSTM(BaseModel):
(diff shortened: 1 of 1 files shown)


googlehydrology/modelzoo/mean_embedding_forecast_lstm.py
```diff
@@ -295,30 +306,36 @@ def _calc_head(self, forecast: torch.Tensor) -> dict[str, torch.Tensor]:
 @dataclasses.dataclass(frozen=True, kw_only=True)
 class ConfigData:
     @classmethod
-    def from_config(cls, cfg: Config) -> "ConfigData":
+    def from_config(cls, cfg: Config) -> 'ConfigData':
+        hindcast_inputs_grouped = group_by_prefix(cfg.hindcast_inputs)
+        forecast_inputs_grouped = group_by_prefix(cfg.forecast_inputs)
+        shared_groups = list(
```
amitmarkel, line 312: No need to convert to list here.

Should convert to lists if needed below, or, just change the type to e.g. `Sequence` IMHO. Let's make the typing work for us instead of vice versa.

---

Also note that list is mutable. We should use Sequence regardless (or tuple)

googlehydrology/modelzoo/mean_embedding_forecast_lstm.py
```diff
@@ -295,30 +306,36 @@ def _calc_head(self, forecast: torch.Tensor) -> dict[str, torch.Tensor]:
 @dataclasses.dataclass(frozen=True, kw_only=True)
 class ConfigData:
     @classmethod
-    def from_config(cls, cfg: Config) -> "ConfigData":
+    def from_config(cls, cfg: Config) -> 'ConfigData':
+        hindcast_inputs_grouped = group_by_prefix(cfg.hindcast_inputs)
```
amitmarkel, line 310: Should convert to set once at the beginning to simplify below

googlehydrology/modelzoo/mean_embedding_forecast_lstm.py
```diff
@@ -23,6 +23,7 @@
 from googlehydrology.modelzoo.fc import FC
 from googlehydrology.modelzoo.head import get_head
 from googlehydrology.utils.config import Config
+from googlehydrology.utils.configutils import group_by_prefix
```
amitmarkel, line 26: Sanity check: no other files were changed like the yaml or configutils?

amitmarkel (approved):
Thanks Omri

googlehydrology/modelzoo/mean_embedding_forecast_lstm.py
omrishefi: No need.

googlehydrology/modelzoo/mean_embedding_forecast_lstm.py
omrishefi: It's different data structures.

googlehydrology/modelzoo/mean_embedding_forecast_lstm.py
omrishefi: Let's do these fixes in next PRs.

googlehydrology/modelzoo/mean_embedding_forecast_lstm.py
amitmarkel: They're the same in practice. It's just to match typing, making it complicated than needed

PR #62, opened 2025-10-30
Configure FC networks with a dynamic embedding size

omrishefi:


googlehydrology/modelzoo/mean_embedding_forecast_lstm.py
```diff
@@ -307,6 +308,13 @@ def _calc_head(self, forecast: torch.Tensor) -> dict[str, torch.Tensor]:
 class ConfigData:
     @classmethod
     def from_config(cls, cfg: Config) -> 'ConfigData':
+        statics_embedding = cfg.statics_embedding
```
amitmarkel: Should couple definition and assert to prevent forgetting to assert a new value in the future etc - more robust against coding errors

`assert (statics_embedding := cfg.statics_embedding) is not None`

etc

googlehydrology/modelzoo/mean_embedding_forecast_lstm.py
```diff
@@ -323,15 +331,19 @@ def from_config(cls, cfg: Config) -> 'ConfigData':
...
             static_attributes_names=tuple(cfg.static_attributes),
             hindcast_inputs_grouped=hindcast_inputs_grouped,
             forecast_inputs_grouped=forecast_inputs_grouped,
             shared_groups=shared_groups,
         )
 
     hidden_size: int
-    embedding_size: int
+    statics_embedding: dict
+    hindcast_embedding: dict
```
amitmarkel, line 347: Just a `dict` type isn't safe right? The code assumes `'hiddens'` earlier. May define it like so for example:

```
from typing import TypedDict

class GoodName(TypedDict):
    hiddens: list[int]
    # etc if another field is needed
```

omrishefi: It's not the matter of this PR. In next PR, the dict will be turned into a structured object. This PR takes the embedding size value from the configuration.

amitmarkel: It's a bad type

googlehydrology/modelzoo/mean_embedding_forecast_lstm.py
omrishefi: Done.

googlehydrology/modelzoo/mean_embedding_forecast_lstm.py
omrishefi: I agree, but changing it in this PR makes it much more complicated. This PR kept the same type which was used before in the library. It will be handled in next PR.

amitmarkel: Uty though, I can fix this in a PR if you'd like. It's technical debt to leave it

amitmarkel (approved):

googlehydrology/modelzoo/mean_embedding_forecast_lstm.py
omrishefi: Please don't. I'm handling it.

PR #63, opened 2025-10-30
Setup EmbeddingSpec type and validation

amitmarkel:


googlehydrology/utils/config.py
```diff
@@ -33,6 +34,12 @@ class WeightInitOpt(str, enum.Enum):
     LSTM_IH_XAVIER = "lstm-ih-xavier"
     FC_XAVIER = "fc-xavier"
 
+class EmbeddingsDict(TypedDict):
```
omrishefi: It should be called EmbeddingSpec

omrishefi: Why not using dataclass? I don't see a reason to keep it as a dictionary.

googlehydrology/modelzoo/mean_embedding_forecast_lstm.py
```diff
@@ -344,8 +344,8 @@ def from_config(cls, cfg: Config) -> 'ConfigData':
 
     hidden_size: int
     statics_embedding: dict
```
omrishefi: What about switching the statics_embedding?

googlehydrology/utils/config.py
```diff
@@ -33,6 +34,12 @@ class WeightInitOpt(str, enum.Enum):
...
+    hiddens: list[int]
```
omrishefi, line 40: I didn't see any change in the access to the fields, i.e. it's still dict['hiddens']. I think it can be much more readable if we access it as properties, i.e. embeddings.hiddens

omrishefi (commented):
See comments

googlehydrology/modelzoo/mean_embedding_forecast_lstm.py
amitmarkel: Done

googlehydrology/utils/config.py
amitmarkel: Done

googlehydrology/utils/config.py
amitmarkel: There was no need: all statically type checked. I verified this. This PR's intent wasn't to do a refactor all over the place but to add typing validation.

Note dataclass on its own doesn't validate anything.

However, changed to dataclass as I agree it's more readable. Also moved validation to the type itself to ensure it's always validated when instantialized. Intent was to do this in a future PR.

googlehydrology/utils/config.py
amitmarkel: Because it's a refactor. Intent was to just validate types first. Also note dataclass on its own is insufficient, it doesn't validate. Intent was just to add validation. Anyway I did change dataclass, but intent was to do this in separate in a future PR.

omrishefi (approved):

PR #68, opened 2025-11-02
Wrapping the networks with ModuleDict

omrishefi:
This might fix the failure with compilation of the networks, but not necessarily.


amitmarkel:
Why is it needed now and not before? A description would've been really helpful. Thanks.

amitmarkel (approved):

PR #71, opened 2025-11-02
Support features lists as different types

omrishefi:
Allow features inputs from config (hindcast, forecast) to be wither list[str], list[list[str]] or dict[str, list[str]]. This provides flexibility for clients to group feature lists by their group names.

### DIFF
--- a/googlehydrology/utils/configutils.py
+++ b/googlehydrology/utils/configutils.py
@@ -68,13 +68,23 @@ def flatten_feature_list(data: list[str] | list[list[str]] | dict[str, list[str]
     return list(data)
 
 
-def group_by_prefix(features: Iterable[str]) -> dict[str, set[str]]:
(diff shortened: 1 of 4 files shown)


googlehydrology/utils/configutils.py
```diff
@@ -68,13 +68,23 @@ def flatten_feature_list(data: list[str] | list[list[str]] | dict[str, list[str]
...
+    For lists, assumes the prefix is written with an underscore for the first feature name.
     """
-    result = {}
-    for feature in features:
-        prefix = feature.partition('_')[0]
-        result.setdefault(prefix, set()).add(feature)
-    return result
+    if not features:
+        return {}
+    if isinstance(features, dict):
```
amitmarkel, line 79: IMHO we should define a dataclass via pydantic to assert and parse. E.g. this features[0] isn't safe etc.

amitmarkel (dismissed):

amitmarkel (approved):

PR #79, opened 2025-11-03
Validate all basins by default.

amitmarkel:


config/multimet_mean_embedding_forecast_lstm.yml
```diff
@@ -41,7 +41,7 @@ metrics:
 num_workers: 0  # Scalene profiler works only without multiprocessing (`0`)
 optimizer: Adam
 validate_every: 2
-validate_n_random_basins: 300
+validate_n_random_basins: 500
```
omrishefi: Let's make it all basins, can use 1000 to take all of them

amitmarkel: Done

config/multimet_mean_embedding_forecast_lstm.yml
```diff
@@ -41,7 +41,7 @@ metrics:
 num_workers: 0  # Scalene profiler works only without multiprocessing (`0`)
 optimizer: Adam
 validate_every: 2
-validate_n_random_basins: 300
+validate_n_random_basins: -1
```
omrishefi, line 44: Please validate that it works well

amitmarkel: 1. Handoff uses it and all previous configs in the library, all used the explicit `-1`.
2. We already validated it but I've just validated it again and it works (`'validate_n_random_basins' not set or set to zero. Will validate on the entire validation set.`)
    
jpg

3. https://github.com/google-research/flood-forecasting/blob/7399428f15689c3f6a9fe5b66233a2530d894538/googlehydrology/training/basetrainer.py#L205
4. Another option is to have it "unset" instead of `-1` but `-1` is more explicit and used by all other configs. I don't mind though here to use ε (empty).

omrishefi (dismissed):

amitmarkel:
Had to merge from main which requires review...

omrishefi (approved):

PR #92, opened 2025-11-05
Use lead_time zero when calculating metrics

omrishefi:


amitmarkel:
Shouldn't this be configurable though? (First or last lead time)

omrishefi:
I'm not sure it will be necessary as config. It complicated the calcualtion. For now, this should be consistent.

amitmarkel (approved):
The slice calculation `slice(-predict_last_n[freq], -predict_last_n[freq] + 1)` should be in a helper variable or method

PR #95, opened 2025-11-06
Remove umal and gmm.

amitmarkel:
Tested [mean_embedding,handoff]x[train,infer] works


docs/source/usage/config.rst
```diff
@@ -142,7 +142,7 @@ General model configuration
    e.g., [``cudalstm``, ``ealstm``, ``mtslstm``]
 
 -  ``head``: The prediction head that is used on top of the output of
-   the model class. Currently supported are ``regression``, ``gmm``, ``cmal``, ``cmal_deterministic`` and ``umal``.
```
omrishefi, line 145: Consider also removing cmal_deteministic to simplify things.

amitmarkel: IMHO cmal_deterministic can be helpful and resembles the global model that uses it if someone wants to use it

omrishefi (approved):

docs/source/usage/config.rst
omrishefi: Please verify it works well. From what I remember it had some bugs, no?

amitmarkel: It was more susceptible to bad data, but that could be just the several other bugs we've just fixed.

I did testing on it before adding with tests both visually and statistically on several tenths of basins data.

---

I'm not against further testing and fixing if needed.

PR #126, opened 2025-11-09
Fixing test test_forecast_daily_regression

omrishefi:
Reading the zarr file instead of a pickle file

### DIFF
--- a/test/test_config_runs.py
+++ b/test/test_config_runs.py
@@ -14,14 +14,16 @@
 
 """Integration tests that perform full runs."""
 
+
 import pickle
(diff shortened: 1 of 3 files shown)


test/test_config_runs.py
```diff
@@ -14,14 +14,16 @@
 
 """Integration tests that perform full runs."""
 
+
 import pickle
```
amitmarkel: Let's remove import pickle

amitmarkel (dismissed):

amitmarkel (approved):

PR #141, opened 2025-11-09
Optimize `normalized_zero` creation in _handle_negative_values.

amitmarkel:
1. Create it directly on GPU instead of on CPU with memory transfers.
2. Do the negation on the in-memory item directly, not on the entire tensor after creation.

### DIFF
--- a/googlehydrology/utils/samplingutils.py
+++ b/googlehydrology/utils/samplingutils.py
@@ -128,7 +128,9 @@ def _handle_negative_values(
         .sel(parameter='scale')
         .item()
     )
-    normalized_zero = torch.tensor(-center / scale, device=values.device)
+    normalized_zero = torch.tensor(
+        -center / scale, device=values.device, dtype=values.dtype
+    )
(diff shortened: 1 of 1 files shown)


googlehydrology/utils/samplingutils.py
```diff
@@ -128,7 +128,9 @@ def _handle_negative_values(
         .sel(parameter='scale')
         .item()
     )
-    normalized_zero = -torch.tensor(center / scale).to(values)
+    normalized_zero = torch.tensor(
+        -center / scale, device=values.device, dtype=values.dtype
```
omrishefi: I think writing -(center/scale) is more readable

omrishefi (dismissed):

googlehydrology/utils/samplingutils.py
amitmarkel: Done.

amitmarkel:
(need reapproval due to merging with main)

omrishefi (approved):

PR #149, opened 2025-11-11
Fix test_forecast_daily_regression() test to run with new model

omrishefi:
Fix test_forecast_daily_regression() test to run with mean_embedding_forecast_lstm.

### DIFF
--- a/googlehydrology/modelzoo/mean_embedding_forecast_lstm.py
+++ b/googlehydrology/modelzoo/mean_embedding_forecast_lstm.py
@@ -327,15 +327,12 @@ class MeanEmbeddingForecastLSTM(BaseModel):
 class ConfigData:
     @classmethod
     def from_config(cls, cfg: Config) -> 'ConfigData':
-        assert (statics_embedding := cfg.statics_embedding) is not None
-        assert (
-            hindcast_embedding := cfg.hindcast_embedding
-            or cfg.dynamics_embedding
-        ) is not None
-        assert (
-            forecast_embedding := cfg.forecast_embedding
-            or cfg.dynamics_embedding
-        ) is not None
+        statics_embedding = cfg.statics_embedding
+        hindcast_embedding = cfg.hindcast_embedding or cfg.dynamics_embedding
+        forecast_embedding = cfg.forecast_embedding or cfg.dynamics_embedding
+        assert statics_embedding is not None
+        assert hindcast_embedding is not None
+        assert forecast_embedding is not None
 
         hindcast_inputs_grouped = group_features_list(cfg.hindcast_inputs)
         forecast_inputs_grouped = group_features_list(cfg.forecast_inputs)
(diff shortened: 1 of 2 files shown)


googlehydrology/modelzoo/mean_embedding_forecast_lstm.py
```diff
@@ -328,8 +328,14 @@ class ConfigData:
     @classmethod
     def from_config(cls, cfg: Config) -> 'ConfigData':
         assert (statics_embedding := cfg.statics_embedding) is not None
-        assert (hindcast_embedding := cfg.hindcast_embedding) is not None
-        assert (forecast_embedding := cfg.forecast_embedding) is not None
+        assert (
```
amitmarkel: Should separate into definition and verification as this isn't a one-liner anymore:

statics..=...
assert statics...

hindcast..=..
assert ..

...

---

Another option is to drop the `is not None`, it shouldn't be empty either, right?

test/conftest.py
```diff
@@ -76,6 +76,16 @@ def _forecast_config_updates(forecast_model):
         if forecast_model.lower() == 'handoff_forecast_lstm':
             update_dict['forecast_overlap'] = 10
             update_dict['regularization'] = ['forecast_overlap']
+        if forecast_model.lower() == 'mean_embedding_forecast_lstm':
+            update_dict['forecast_overlap'] = 30
+            update_dict['loss'] = 'CMALLoss'
+            update_dict['head'] = 'cmal'
```
amitmarkel: The name of the test is regression so should use regression no?

I'm in favor of running both heads

omrishefi: CMAL should be a different test. Removed these lines for now.

googlehydrology/modelzoo/mean_embedding_forecast_lstm.py
```diff
@@ -327,9 +327,12 @@ def _calc_head(
 class ConfigData:
     @classmethod
     def from_config(cls, cfg: Config) -> 'ConfigData':
-        assert (statics_embedding := cfg.statics_embedding) is not None
-        assert (hindcast_embedding := cfg.hindcast_embedding) is not None
-        assert (forecast_embedding := cfg.forecast_embedding) is not None
+        statics_embedding = cfg.statics_embedding
+        hindcast_embedding = cfg.hindcast_embedding or cfg.dynamics_embedding
```
amitmarkel, line 331: Noting this changes the logic of the model, this pr is about fixing the test, no?

amitmarkel (approved):

googlehydrology/modelzoo/mean_embedding_forecast_lstm.py
omrishefi: It only adds a fallback for the embedding networks spec. This isn't used in the yml we provide. It is necessary for making the test pass.

PR #152, opened 2025-11-13
Remove obsolete unit tests

omrishefi:
- Remove yml files related to old models.
- Remove tests not relevant for forecast.
- Add uncertainty test for mean_embedding_forecast_lstm.
- Handoff model doesn't support CMAL head.
- CMALL deteministic doesn't support negative_sample_handling=clip


amitmarkel:
PR SG

About cmal deterministic. It doesn't call `_handle_negative_values()` so it doesn't apply negative_sample_handling=clip" but nor applies =truncate. "Doesn't support" means it fails if set and/or fails when testing for clipping?

Anyway, I didn't optimize "truncate". We always ran with "clip". So:

1. Let's add a TODO on cmal deterministic to call/support _handle_negative_values().
2. Let's keep setting the value to "clip", not "truncate" or none value.

amitmarkel (approved):

PR #155, opened 2025-11-16
[optimize] Detach scaler check zero and save from the graph

amitmarkel:
Dask needs to serialize the parent subgraph upwards to send args to Delayed function, even if the leaf node is very small.

So, we need to validate not as part of dask's computation graph. Made computation a part of the main compute() so it's efficient.

Runtime now is practically immediate, e.g. for 3000 basins over 40yrs: 0.00039s instead of 5.25m (~800,000x faster)

Another option may be to make a division by zero inherent error when a scale is zero, but it won't be descriptive and confusing, and it would need to be exposed to multimet's compute() as well regardless.

### DIFF
--- a/googlehydrology/datautils/scaler.py
+++ b/googlehydrology/datautils/scaler.py
@@ -18,8 +18,6 @@ from typing import Iterator, Hashable
 
 import dask
 import dask.array
-import dask.graph_manipulation
-import dask.delayed
 import pandas as pd
 import xarray as xr
 
@@ -103,7 +101,7 @@ class Scaler:
         self.scaler_dir = scaler_dir
         if not calculate_scaler:
             self.load()
-            self._check_zero_scale()
+            self.check_zero_scale()
         else:
             self._custom_normalization = custom_normalization
             if dataset is not None:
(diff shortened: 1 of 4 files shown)


googlehydrology/datautils/scaler.py
```diff
@@ -166,17 +162,34 @@ def calculate(
...
+        assert not chunks, f'`scaler` needs to be computed yet has {chunks=}'
 
-    def _check_zero_scale(self):
-        _check_zero_scale(self.scaler)
+        os.makedirs(self.scaler_dir, exist_ok=True)
+        scaler_file = self.scaler_dir / SCALER_FILE_NAME
+        with open(scaler_file, 'wb') as f:
+            self.scaler.to_netcdf(f)
+
+    def _prepare_check_zero_scale_input(self):
```
omrishefi: Let's document this split, and explain why we did it. We should also evaluate what is the impact of merging the two functions together. scaler dataset should be very minial in it's size so the copy not a large memory consumption.

omrishefi (commented):
Please see comment

googlehydrology/datautils/scaler.py
amitmarkel: Done

omrishefi (approved):

PR #158, opened 2025-11-17
Allow disabling disk swap for workers.

amitmarkel:
If we want memory laziness etc, it should be via the OS +/ our lazy solution or other solution such as loading only the relevant dataset's subset into memory.

### DIFF
--- a/googlehydrology/run.py
+++ b/googlehydrology/run.py
@@ -88,14 +88,14 @@ def _main():
 
     torch.autograd.set_detect_anomaly(config.detect_anomaly)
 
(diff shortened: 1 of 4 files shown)


googlehydrology/run.py
```diff
@@ -90,6 +90,7 @@ def _main():
 
     dask.config.set(
         {
+            'distributed.p2p.storage.disk': False,
```
omrishefi: Let's add config

omrishefi (dismissed):

omrishefi (approved):

PR #170, opened 2025-11-23
Add load_as_csv:False

amitmarkel:


config/multimet_mean_embedding_forecast_lstm.yml
```diff
@@ -96,6 +96,7 @@ targets_data_dir: ~/nh-data/data/Caravan-nc
 statics_data_dir: ~/nh-data/data/Caravan-nc
 dynamics_data_dir: ~/nh-data/data/Caravan-multimet
 
+load_as_csv: False
```
omrishefi, line 99: Isn't it the default?

omrishefi (commented):
see my comment

config/multimet_mean_embedding_forecast_lstm.yml
amitmarkel: Yes. Can have it with empty value in new PR, I can't at the moment to. Clearer when set though.

omrishefi (approved):

PR #175, opened 2025-12-27
Removed unused code -- dynamic featers & basin ID encoding

grey-nearing:
Removed unused code of two types:

1) Old `dynamic_features`, which has been replaced with `forecast_features` and `hindcast_features`. We no longer support non-forecast models.

2) Basin ID encodings, which would be nice to have but is not currently supported.


googlehydrology/datasetzoo/multimet.py
```diff
@@ -105,16 +105,8 @@ def __init__(
         self._static_features = cfg.static_attributes
         self._target_features = cfg.target_variables
         self._forecast_features = []
```
amitmarkel: Missed leftover (line 107) to remove (empty list prior to now-removed `if`).

googlehydrology/datautils/validate_samples.py
```diff
@@ -21,89 +21,6 @@
 LOGGER = logging.getLogger(__name__)
 
 
-def _flatten_feature_groups(groups: list[list[str]]) -> list[str]:
```
amitmarkel, line 24: Please check unit tests for all modules. Good chance to need to fix tests or remove.

amitmarkel (changes requested):

amitmarkel:
Btw as always please run a very short but complete: 2 epocs train+valid and infer - for each PR

amitmarkel:
Screenshot 2025-12-28 at 15 32 16

BTW on PRs, please rerequest review so I get an email and also know when to rereview.

PR #179, opened 2025-12-27
Clean config

grey-nearing:
Removed unused config arguments.


googlehydrology/utils/config.py
```diff
@@ -193,9 +184,7 @@ def dump_config(self, folder: Path, filename: str = 'config.yml'):
                                 )
                             temp_cfg[key] = temp_list
                         else:
-                            # Ignore None's due to e.g. using a per_basin_period_file
-                            if isinstance(val, pd.Timestamp):
```
amitmarkel, line 197: Why is `if isinstance(val, pd.Timestamp):` deleted?

If it should so always indeed, then please add an `assert isinstance(val, pd.Timestamp)`

amitmarkel (changes requested):

PR #180, opened 2025-12-27
Clean loss

grey-nearing:
Removed some unused loss and regularization components, including the tie-frequencies regularization and the (depreciated) weightedNSE loss.


googlehydrology/training/__init__.py
```diff
@@ -105,12 +103,6 @@ def get_loss_obj(cfg: Config) -> loss.BaseLoss:
         loss_obj = loss.MaskedMSELoss(cfg)
     elif cfg.loss.lower() == 'nse':
         loss_obj = loss.MaskedNSELoss(cfg)
-    elif cfg.loss.lower() == 'weightednse':
```
amitmarkel, line 108: Reflected in documentation?

amitmarkel (approved):

googlehydrology/training/__init__.py
grey-nearing: It was still in multimet.py and the multimet test. I've removed it from both.

PR #183, opened 2025-12-27
Remove cmal deterministic config arg

grey-nearing:
The codebase determines whether to use deterministic CMAL sampling by checking model.cfg.head.lower() == 'cmal_deterministic' (see utils/samplingutils.py). It does not check this boolean configuration flag.

The Multimet dataset class (in datasetzoo/multimet.py) manually imports and calls the utility function flatten_feature_list(cfg.forecast_inputs) instead of accessing forecast_inputs_flattened convenience property. Same with hindcast.

### DIFF
--- a/googlehydrology/utils/config.py
+++ b/googlehydrology/utils/config.py
@@ -499,10 +499,6 @@ class Config(object):
         return self._get_value_verbose('head')
 
     @property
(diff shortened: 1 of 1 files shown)


amitmarkel (changes requested):
SG but let's merge first

PR #184, opened 2025-12-27
Remove plots

grey-nearing:
Removed all plotting functionality. This can be done by the user during data analysis and is otherwise code that must be maintained.


amitmarkel (changes requested):
Let's sync with Omri on this one.

grey-nearing:
> Let's sync with Omri on this one.

Ok, I'm skipping this one for now. We can revisit later.

amitmarkel:
> > Let's sync with Omri on this one.
> 
> Ok, I'm skipping this one for now. We can revisit later.

This change does SG but Omri said IIRC that those were helpful. So I suggest to still keep the change to revisit later.

amitmarkel:
Sync'd with Omri: also in favour of keeping plots

PR #190, opened 2025-12-30
Created Example Configs for FloodHub Approximation

grey-nearing:
This commit includes changes to the Handoff Forecast LSTM model, because that model was not suited to replicate the Nature paper and former Google FloodHub settings. 

This commit also removes the old example configs directory, since the current one supersedes. 

It removes the Input Layer altogether, since it is no longer used anywhere.

Finally, it updates docs accordingly.

It appears that no unit tests need updating.


config/basins/caravan/camels.txt
```diff
@@ -1,670 +0,0 @@
-camels_01013500
```
amitmarkel, line 1: I'm using this file. Potentially also the other files in this directory for testing still. Let's remove them at a later time.

example-configs/floodhub-settings-config.yml
```diff
@@ -0,0 +1,383 @@
...
+
+# If True, samples with all-zero inputs are treated as invalid data gaps.
+allzero_samples_are_invalid: False
+
+
+# --- 6. System & Runtime ------------------------------------------------------
+# Computational settings.
+
+# Number of CPU processes used to load data in parallel. Higher can be faster.
+num_workers: 4
```
amitmarkel: Suggesting this to be 0.

Without further optimizations I'm working on, this currently results in multiplying memory usage by `num_workers + 1`.

amitmarkel (approved):
Please ensure both models train+validate+infer as expected

config/basins/caravan/camels.txt
grey-nearing: Ok, understood. Let's merge this deletion now and break the test. The unit tests are broken anyway currently. 

 I will submit a new PR with the unit tests fixed presently, including this fix. 
 
 Going forward, it would be good to have more test coverage, and also to integrate the testing into the github merge pipeline.

example-configs/floodhub-settings-config.yml
grey-nearing: This is very good to know. Probably why I am having memory issues in my hypertuning runs. Thanks.

PR #191, opened 2025-12-31
Update existing unit tests

grey-nearing:
Several unit tests were broken based on recent changes. The following tests are fixed:

1) `test_config_runs.py` --> removed unused config args and updated the call to caravan's `load_caravan_timeseries` (now called `load_caravan_timeseries_together`). 

2) `test_multimet.py` --> fixed expected error messages due to removal of dynamic_imports as a config option. 

3) `test_validate_samples.py` --> removed references to functions that were removed as part of the cleanup.

The `test_uncertainty.py` was NOT updated in this PR and is still broken. All other tests work.

Future work should increase test coverage beyond the current 71%.

### DIFF
--- a/test/test_config_runs.py
+++ b/test/test_config_runs.py
@@ -90,9 +90,9 @@ def _check_results(config: Config, basin: str, discharge: pd.Series = None):
     )
 
     if discharge is None:
-        discharge = caravan.load_caravan_timeseries(
-            config.data_dir, basin
-        ).to_dataframe()[config.target_variables]
+        discharge = caravan.load_caravan_timeseries_together(
+            config.data_dir, basin, config.target_variables, csv=False
+        ).to_dataframe()
 
     if hasattr(config, 'lead_time'):
         results = results.isel(time_step=0).squeeze()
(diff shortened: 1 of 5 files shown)


test/test_config_runs.py
```diff
@@ -90,19 +98,23 @@ def _check_results(config: Config, basin: str, discharge: pd.Series = None):
...
 
     results_array = results[f'{config.target_variables[0]}_obs'].values
-    discharge_array = np.array(
-        [v.item() for v in discharge.loc[test_start_date:test_end_date].values]
-    )
+    idx = pd.IndexSlice
+    try:
+        discharge_slice = discharge.loc[idx[basin, test_start_date:test_end_date], 'streamflow']
+    except:
+        import pdb; pdb.set_trace()
```
amitmarkel: Should remove the try-except with pdb

amitmarkel (approved):
Thanks Grey

test/test_config_runs.py
grey-nearing: I always do this ... It's removed.

PR #196, opened 2026-01-03
Fixed scaler test and updated conda env for h5netcdf.

grey-nearing:
Conda environment needs h5netcdf after recent changes. Also, one of the unit tests needed to be updated to explicitly save the scaler with h5netcdf.


amitmarkel (approved):
Nice catch

PR #195, opened 2026-01-03
Updated documentation.

grey-nearing:
This is a merge into main of a stack of PRs. The purpose of these PRs was to update the user-facing documentation for the repository, including adding a tutorial and adding a set of example config files for both models that are currently supported. Updating the documentation required first removing all of the unused code that remained in the repository from when it was forked from NeuralHydrology.

**Related outstanding tasks:**

0) Remove frequency-related code throughout the repository.
1) Complete test coverage, including fixing `test_uncertainty.py`, and integrating testing into the github PR workflow.
2) Activating ReadTheDocs when the repo goes public.

**PRs represented in this commit are:**

 https://github.com/google-research/flood-forecasting/pull/175

https://github.com/google-research/flood-forecasting/pull/176

https://github.com/google-research/flood-forecasting/pull/177

https://github.com/google-research/flood-forecasting/pull/178

https://github.com/google-research/flood-forecasting/pull/179

https://github.com/google-research/flood-forecasting/pull/180

https://github.com/google-research/flood-forecasting/pull/181

https://github.com/google-research/flood-forecasting/pull/182

https://github.com/google-research/flood-forecasting/pull/183

184: *NOT INCLUDED* (this is not merged yet).

https://github.com/google-research/flood-forecasting/pull/185

186: *NOT INCLUDED* (this was merged into main prior to this commit).

187: *NOT INCLUDED* (this was merged into main prior to this commit).

188: *NOT INCLUDED* (this was merged into main prior to this commit).

https://github.com/google-research/flood-forecasting/pull/189

https://github.com/google-research/flood-forecasting/pull/190
 
https://github.com/google-research/flood-forecasting/pull/191

PR #197, opened 2026-01-03
Remove Config Directory

grey-nearing:
Removed the old config directory that contains outdated config files. Example config files are now placed in the `example_configs` directory and a small test that the user can run is in the `tutorial` directory. 

Going forward, everything needed to run small tests (configs, basins lists, and data) are in the `tutorial` directory. A small example training run with the `mean_embedding_forecast_lstm` can be run using the config file `~/tutorial/congifs/train-config.yml`:

`~/flood-forecasting$ python googlehydrology/run.py train --config-file=tutorial/configs/train-config.yml`

No unit tests rely on this directory.

Please notice that this PR merges directly to `main`, but contains all changes in https://github.com/google-research/flood-forecasting/pull/195. Please merge PR #195 first.


amitmarkel:
Please merge the other changes.

Also let's keep camels.txt for now.

PR #207, opened 2026-01-06
[optimize memory] make index_sampler dict virtual so it takes no memory.

amitmarkel:
Implement SampleIndexer which creates items on the fly.
This saves 50GB for a dataset period of 46 years over all 16k basins. (Indices take 1.7GB.)

(Note: the dict was efficient to create runtime-wise (1 min).)

I did a verification to assert a computed dict out of SampleIndexer equalled to the created dict.

Noting that `items` may be improved to avoid calling `__getitem__` repeatedly.

---

This PR also splits the dict creation function into two: valid samples mask and indices being returned to init so it can compute() on it in separate, and the dict creation itself.

### DIFF
--- a/googlehydrology/datasetzoo/multimet.py
+++ b/googlehydrology/datasetzoo/multimet.py
@@ -242,9 +242,6 @@ class Multimet(Dataset):
         LOGGER.debug('scale data')
         self._dataset = self.scaler.scale(self._dataset)
 
-        LOGGER.debug('create valid sample mask and indices')
-        valid_sample_mask, indices = self._create_valid_sample_mask()
-
         # TODO: Optionally, optimize the data loader and trainer modules to work with chunked lazy data.
         LOGGER.debug('compute dataset, scaler')
         # We explicitly keep the self.scaler.scaler computation since trainer uses it directly
--- a/googlehydrology/datasetzoo/multimet.py
+++ b/googlehydrology/datasetzoo/multimet.py
@@ -16,6 +16,7 @@ import functools
 import itertools
 import logging
 import math
+import multiprocessing
 import pickle
 import subprocess
 import sys
@@ -241,13 +242,20 @@ class Multimet(Dataset):
         LOGGER.debug('scale data')
         self._dataset = self.scaler.scale(self._dataset)
 
+        LOGGER.debug('create valid sample mask and indices')
+        valid_sample_mask, indices = self._create_valid_sample_mask()
+
         # TODO: Optionally, optimize the data loader and trainer modules to work with chunked lazy data.
         LOGGER.debug('materialize data (compute)')
         # We explicitly keep the self.scaler.scaler computation since trainer uses it directly
-        # create sample index does a compute on the data. We compute here prior to avoid recompute.
-        self._dataset, self.scaler.scaler = dask.compute(
-            self._dataset, self.scaler.scaler
-        )
+        with multiprocessing.Pool(processes=1) as pool:
+            self._dataset, self.scaler.scaler, indices = dask.compute(
+                self._dataset,
+                self.scaler.scaler,
+                indices,
+                pool=pool,
+                scheduler='processes',
+            )
 
         LOGGER.debug('scaler check zero scale')
         self.scaler.check_zero_scale()
@@ -255,8 +263,7 @@ class Multimet(Dataset):
         self.scaler.save()
 
         # Create sample index lookup table for `__getitem__`.
-        LOGGER.debug('create sample index')
-        self._create_sample_index()
+        self._create_sample_index(valid_sample_mask, indices)
 
         # Compute stats for NSE-based loss functions.
         # TODO (future) :: Find a better way to decide whether to calculate these. At least keep a list of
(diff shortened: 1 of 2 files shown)


omrishefi (dismissed):

omrishefi (commented):
Please verify that passing all data between processes is possible

amitmarkel:
Everything is working properly but I'm considering to improve the sample_index dict further first. Its impact is greater than this optimization.

For example, I'm testing a reduction of 4.3GB for 671 basins over 46y to ~155MB which is 80% of memory.

googlehydrology/datasetzoo/multimet.py
```diff
@@ -241,22 +242,29 @@ def __init__(
         LOGGER.debug('scale data')
         self._dataset = self.scaler.scale(self._dataset)
 
+        LOGGER.debug('create valid sample mask and indices')
+        valid_sample_mask, indices = self._create_valid_sample_mask()
```
omrishefi: This is called twice. Is it a leftover of the previous PR?

googlehydrology/datasetzoo/multimet.py
```diff
@@ -475,12 +483,7 @@ def _union_ranges(
         ]
         return functools.reduce(pd.Index.union, ranges)
 
-    # This is run by the base class.
-    def _create_sample_index(self):
-        """Creates a map from integer sample indexes to the integer positions into the xr.Dataset.
-
-        This allows index-based sample retrieval, which is faster than coordinate-based sample retrieval.
-        """
+    def _create_valid_sample_mask(self):
```
omrishefi, line 483: Add the documentation to this function.

googlehydrology/datasetzoo/multimet.py
```diff
@@ -500,13 +503,22 @@ def _create_sample_index(self):
...
         # Convert boolean valid sample mask into indexes of all samples. This retains
         # only the portion of the valid sample mask with True values.
         # Each element is a list of valid integer positions (indexers) for which
         # values are True for a dimension.
-        indices = dask.compute(*dask.array.nonzero(valid_sample_mask.data))
+        indices = dask.array.nonzero(valid_sample_mask.data)
+
+        return valid_sample_mask, indices
 
+    def _create_sample_index(
```
omrishefi, line 516: It's not clear to me if the split of the function to two functions is related to this PR optimization or not.

omrishefi (commented):
Please see my comments

googlehydrology/datasetzoo/multimet.py
amitmarkel: Done (forgot to stage this block change)

googlehydrology/datasetzoo/multimet.py
amitmarkel: Moved to here, and added one below instead.

googlehydrology/datasetzoo/multimet.py
amitmarkel: It's better to have the compute where the other one is, instead of being hidden having a comment it's there.

amitmarkel (commented):
Thanks Omri

googlehydrology/datasetzoo/multimet.py
```diff
@@ -973,3 +973,35 @@ def _get_products_and_bands_from_feature_strings(
             product = 'ERA5_LAND'
         product_bands.setdefault(product, []).append(feature)
     return product_bands
+
+class SampleIndexer:
```
omrishefi, line 977: How can we emphesize this class acts and implements the dictionary functions?

test/test_tester.py
```diff
@@ -13,33 +13,41 @@
...
-        # basin 103: 2 samples, 1 batch
-        13: {'basin': 103, 'date': 1},
-        14: {'basin': 103, 'date': 2},
-    }
+    sample_index = SampleIndexer(  # sample index -> basin metadata
+        (
+            (
+                'basin',
+                [
+                    # 7 samples, 3 batches
```
omrishefi: In the comment, use the previous comment text "basin 101: 7 samples, 3 batches". Same for lines below.

googlehydrology/datasetzoo/multimet.py
amitmarkel: Thing is, it doesn't implement everything a dict implements. If we do so, we'll need to implement a lot of redundant things.

test/test_tester.py
amitmarkel: Done. Also using np.array because that's the true expected/used type, not a python array.

googlehydrology/datasetzoo/multimet.py
```diff
@@ -242,21 +243,25 @@ def __init__(
...
-
+        LOGGER.debug(f'Dataset size: {sizeof(self._dataset) / 1024 / 1024} MB')
         LOGGER.debug('scaler check zero scale')
         self.scaler.check_zero_scale()
         LOGGER.debug('scaler save')
         self.scaler.save()
 
+        LOGGER.debug('create valid sample mask and indices')
+        valid_sample_mask, indices = self._create_valid_sample_mask()
+        (indices,) = dask.compute(indices)
```
omrishefi, line 259: Please write in the PR description that it includes also this compute optimization, which isn't related to the dictionary optimizations.

omrishefi (commented):
See comments below.

googlehydrology/datasetzoo/multimet.py
amitmarkel: Done.

omrishefi (approved):
Great PR!

amitmarkel:
Thanks Omri. For next PRs I'll separate non-binding changes with the main change, such as the compute() separation.

PR #210, opened 2026-01-07
Added CAMELS-US Configuration Files

grey-nearing:
The ``~/flood-forecasting/example-configs`` directory contains reference configuration files in YAML format. These files define the experimental setups for different model architectures and datasets used in Google's production forecasting systems. These files provide stable baselines for model development and evaluation on standard datasets like CAMELS-US.

There is a difference between the CAMELS tests here and the ones we have been using up to this point. These CAMELS tests use the CAMELS-531 (531 basins) dataset instead of the CAMELS-670 (670 basins) dataset. This is important for public-facing release because the community benchmark is typically conducted on the 531 basin set, not the 670 basin set.

Basin list files necessary for running CAMELS-US and Multimet runs are now included.

The `example-configs` directory now supersedes the old `configs` directory, which was removed. 

README and documentation was updated to reflect this change.


example-configs/multimet-basins-list-with-chirps.txt
amitmarkel, line 1: The no-chirps file is 15841 lines long yet with-chirps is 14262 lines long. Sanity check: does it make sense?

README.md
```diff
@@ -119,6 +119,27 @@ Experiments are defined by YAML files. Update the following paths in your config
...
+  * **Model Architecture:** `mean_embedding_forecast_lstm`  
+  * **Dataset:** MultiMet (Global Caravan dataset)  
+  * **Description:** This configuration is designed to replicate the training settings of the current (2025) operational FloodHub model as closely as possible within this open-source framework.  
+* **`handoff-forecast-lstm-config.yml`**  
+  * **Model Architecture:** `handoff_forecast_lstm`  
+  * **Dataset:** MultiMet (Global Caravan dataset)  
+  * **Description:** Provides the settings used for the former operational model. This configuration aligns with the methodology described in the *Nature* (2024) paper for global ungauged flood prediction.  
+* **`camels-multimet-mean-embedding-forecast-lstm-config.yml`**  
+  * **Model Architecture:** `mean_embedding_forecast_lstm`  
+  * **Dataset:** CAMELS-US (671 basins)
```
amitmarkel: It points to the 531.
1. The yml needs to be updated to point to the former camels.txt (670) as an e.g. camels-us-670-basins-list.txt, right?
2. The 670 basin list seems missing

example-configs/camels-multimet-mean-embedding-forecast-lstm-config.yml
```diff
@@ -0,0 +1,272 @@
...
+train_basin_file: ~/flood-forecasting/example-configs/camels-us-531-basins-list.txt
+validation_basin_file: ~/flood-forecasting/example-configs/camels-us-531-basins-list.txt
+test_basin_file: ~/flood-forecasting/example-configs/camels-us-531-basins-list.txt
+
+# Directory containing the target data (streamflow observations).
+targets_data_dir: ~/data/camels-updated-2025
+# Directory containing static catchment attributes (e.g., area, elevation).
+statics_data_dir: ~/data/camels-updated-2025
+# Directory containing dynamic meteorological forcing data (rain, temp, etc.).
+dynamics_data_dir: gs://caravan-multimet/v1.1
```
amitmarkel: Why not a static dynamics_data_dir like camels? It downloads everything on every run... Getting things very slow.

amitmarkel (commented):
Thanks Grey

README.md
grey-nearing: Good catch, I forgot to update the docs after my last commit changing 671 to 531. The difference between 530 vs. 531 and 670 vs. 671 is inconsequential because there is no data in that one basin. 

I think we want to externalize this with CAMELS-531, not CAMELS-671, since that is the more common academic benchmark.

example-configs/camels-multimet-mean-embedding-forecast-lstm-config.yml
grey-nearing: Good plan. This is changed.

README.md
amitmarkel: Alright

amitmarkel (dismissed):

example-configs/multimet-basins-list-with-chirps.txt
grey-nearing: Wow, good catch. This error goes deep into my cross validation experiments. I had the files misnamed in the colab that creates the cross validation splits. Thank you very much for noticing this.

amitmarkel (approved):

PR #212, opened 2026-01-08
[optimize] Reimplement BasinBatchSampler to lower memory and runtime

amitmarkel:
Limit memory on num basins and runtime to sub second for 16k basins.

Binary search via numpy on already sorted computed indexes instead of a dict to store and resort it. So memory spike is some k times num basins. The dict impl was problematic as it iterated on values and small dicts created on the fly for all values: memory and runtime.

For the 671 camels basin list over 46yr: the dict held 622MB. Runtime was hundreds of seconds for 16k basins at each validation epoc.

The computed indices for each samples are in a flat array (column) like [0,0,0,1,2,2,2,2,3,3,....]. So getting left and right edges via numpy: [0,3,4,8,...] [3,4,8,10,...]. Each defines the batch for a basin in that order [0,1,2] (from the 0,0,0 part) hence the range(0,3) when yielding the batch.

Else for the all basins case, we can look at the col offsetted by one and have numpy find the boundary changes ie where there's differing values. The boundaries are those values offsetted right by one because end indices should always be exclusive ie bounds... Then to get the distances/lengths/counts, np.diff returns that result.

Tested for equality exactly for both 671 basins 1950-2025 and for 16k basins 1980-2025, and the unit tests.

### DIFF
--- a/googlehydrology/evaluation/utils.py
+++ b/googlehydrology/evaluation/utils.py
@@ -97,7 +97,7 @@ class BasinBatchSampler(BatchSampler):
             # Counts are ends - starts ie the distance (diff) eg [3,1,4,2]
             self._starts, self._counts = bounds[:-1], np.diff(bounds)
         # Array lengths are in terms of num basins eg len changes, len starts,
-        # etc. Or use bool masks that are 1 byte instead of 8 bytes (64 bit) so
+        # etc. Or bool masks that are 1 byte instead of 8 bytes (64 bit) so
         # the col's compare size is 1/8 of col.
 
         fulls, remainders = np.divmod(self._counts, batch_size)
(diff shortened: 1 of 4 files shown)


googlehydrology/evaluation/utils.py
```diff
@@ -71,32 +73,41 @@ class BasinBatchSampler(BatchSampler):
...
         batch_size: int,
-        basins_indexes: set[int],
+        basins_indexes: np.typing.NDArray[np.integer],
     ):
         super().__init__(
             SequentialSampler(range(len(sample_index))),
             batch_size,
             drop_last=False,
         )
+        self.batch_size = batch_size
```
omrishefi: Make it private

omrishefi (dismissed):

omrishefi (approved):

PR #214, opened 2026-01-12
Add progress bar for dask compute ops.

amitmarkel:
Allows to see how much work is left and how much work was done at a given moment.

Untitled

### DIFF
--- a/googlehydrology/run.py
+++ b/googlehydrology/run.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 import argparse
+import logging
 import os
 import sys
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


googlehydrology/run.py
```diff
@@ -101,6 +102,13 @@ def _main():
     if config.cache.enabled:
         dask.cache.Cache(cachey.Cache(config.cache.byte_limit)).register()
 
+    tqdm.dask.TqdmCallback(
```
omrishefi: Can we put it behind a flag or enable it only when debug is used?

omrishefi (commented):
See comment

googlehydrology/run.py
amitmarkel: Done

omrishefi (approved):

PR #215, opened 2026-01-12
Release freed memory post load data back to the OS to prevent mem fragmentation

amitmarkel:
For the full caravan over 46yr, 28GB -> 5.5GB at this point.

More memory is used when capacity is higher on a machine, so that figure changes with a machine's spec as dask chooses how much memory to use (num workers in a process and by chunk size).

However, the documentation in memory.py explains why do this. It's better to release unused memory explicitly than risk concurrent racing due to allocations' fragmentation.

The code is compatible/portable for every Linux that uses libc, which is effectively everything: Android isn't a target to support, but anyway a library like musl or bionic libc instead of libc seems not to have trim anyway. In such cases, trimming just won't take place.

### DIFF
--- a/googlehydrology/datasetzoo/multimet.py
+++ b/googlehydrology/datasetzoo/multimet.py
@@ -158,6 +158,7 @@ class Multimet(Dataset):
         # Load & preprocess the data.
         LOGGER.debug('load data')
         self._dataset = self._load_data()
+        memory.release()
         LOGGER.debug('validate all floats are float32')
         _assert_floats_are_float32(self._dataset)
 
--- a/googlehydrology/utils/memory.py
+++ b/googlehydrology/utils/memory.py
@@ -3,6 +3,7 @@
 import contextlib
 import ctypes
 import gc
+from ctypes.util import find_library
 
 
 def release() -> None:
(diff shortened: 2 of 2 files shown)


googlehydrology/datasetzoo/multimet.py
```diff
@@ -158,6 +158,7 @@ def __init__(
         # Load & preprocess the data.
         LOGGER.debug('load data')
         self._dataset = self._load_data()
+        memory.release()
```
omrishefi, line 161: Don't you need to add an import?

googlehydrology/utils/memory.py
```diff
@@ -0,0 +1,27 @@
...
+       memory that isn't used yet because it's fragmented
+       and it was too fast to get reused.
+
+    Linux supports trim. MacOS and Windows release freed
+    memory more aggressively automatically.
+    """
+    gc.collect()
+
+    with contextlib.suppress(OSError, AttributeError):  # Ignore non Unix-like
+        ctypes.CDLL(find_library('c') or 'libc.so.6').malloc_trim(0)
```
omrishefi, line 27: Will it work in windows?

omrishefi (commented):
See comments

googlehydrology/datasetzoo/multimet.py
amitmarkel: It's the underlying c lib for the basic things as mallocs, opening files etc. It's not a python library. It's what the python binary and other binaries must use. 

The find_library part makes it even more maintainable for the future if libc turns to version 3 but that isn't expected for another decade or even two. It's a good practice though in case there's some ultra specialized Linux distribution like an Arduino or something. It's a tiny addition so there wasn't a reason not to add that.

googlehydrology/utils/memory.py
amitmarkel: As documented, windows and macos etc do this more aggressively and don't implement trimming anyway.

It'll work since we skip the code if it errs.

omrishefi (approved):

PR #222, opened 2026-01-14
[optimize memory, runtime] load_caravan_timeseries_together (96% less memory 26gb->1gb, 37% faster 8m->5m)

amitmarkel:
Optimize 26gb mem over 8min --> 1gb mem over 5min. (16k basins, 46yrs). It's 96% less memory and 37% faster.

`open_mfdataset` actually isn't a C function, isn't optimized for our case. It's general. Combines data with high memory use: 26gb (16k basins over 46yrs).

Looked at its args. Main issue is `join`. It needs to align all 16k data date ranges at once with the general method. However can't just use `join='override'` with open_mfdataset since it would just stack basin date ranges without align.

So, align every 500 basins into a batch that's combined with join 'override' since it's aligned. Now got just 32 combined datasets to combine (for 16k basins) with `join='outer'` which finishes in 100ms and takes ~O(1) memory (some 50 MB or so).

What align means: every basin has data in different parts of years (e.g. a basin on 1980-2020 but another on 1990-2016). So each batch we open all basin datasets and get the first and last dates, so we know min and max of all of those. Now reindex each basin to account for nans at every place where there's a mismatch, so everything is aligned, which is what join='outer' results in.

It could be possible to do something else for the final merge but it's so fast and non memory consuming already, it's better not to micro optimize.

Validated it's the same via `xarray.identical()`. Memory use metered by `psutil.Process(os.getpid()).memory_info().rss/1024**3`.

A side effect is we may progress bars for this phase now.

### DIFF
--- a/googlehydrology/datasetzoo/caravan.py
+++ b/googlehydrology/datasetzoo/caravan.py
@@ -12,8 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import functools
 import itertools
 import logging
+import math
+from collections.abc import Iterable, Iterator
 from pathlib import Path
 
 import dask
@@ -22,6 +25,8 @@ import dask.delayed
 import numpy as np
 import pandas as pd
 import xarray
+
+from googlehydrology.utils.tqdm import AutoRefreshTqdm as tqdm
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


googlehydrology/datasetzoo/caravan.py
```diff
@@ -144,13 +149,23 @@ def load_csvs_as_ds(basin_to_path: dict[str, Path]) -> xarray.Dataset:
     datas = [df.assign(basin=basin) for basin, df in zip(basin_to_path, datas)]
     return dd.concat(datas).compute().set_index(['basin', 'date']).to_xarray()
 
+_JOIN = functools.partial(
```
omrishefi: Move to function

omrishefi (commented):
see comment

googlehydrology/datasetzoo/caravan.py
amitmarkel: Done

amitmarkel:
* Moved constants to the function
* Decided to split into another method so each one does only one thing
* Re-validated on [671, 5k, 16k] basins

omrishefi (approved):

PR #230, opened 2026-02-11
[optimize memory] Compactify indices memory width.

amitmarkel:
For example, hundreds of millions up to billions of int64 sample indexes can be stored as uint16 to uint32. So this saves several GBs of memory.

Made sure results' values were the same for several thousands of basins over 40yr.

### DIFF
--- a/googlehydrology/datasetzoo/multimet.py
+++ b/googlehydrology/datasetzoo/multimet.py
@@ -584,7 +584,7 @@ class Multimet(Dataset):
         # values are True for a dimension.
         indices = dask.array.nonzero(valid_sample_mask.data)
         # Compact memory widths. Values are indexes within each mask's shape.
-        dtypes = [np.min_scalar_type(n - 1) for n in valid_sample_mask.shape]
+        dtypes = map(np.min_scalar_type, valid_sample_mask.shape)
         indices = tuple(idx.astype(dt) for idx, dt in zip(indices, dtypes))
 
         return valid_sample_mask, indices
--- a/googlehydrology/datasetzoo/multimet.py
+++ b/googlehydrology/datasetzoo/multimet.py
@@ -583,6 +583,9 @@ class Multimet(Dataset):
         # Each element is a list of valid integer positions (indexers) for which
         # values are True for a dimension.
         indices = dask.array.nonzero(valid_sample_mask.data)
+        # Compact memory widths. Values are indexes within each mask's shape.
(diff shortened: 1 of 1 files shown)


googlehydrology/datasetzoo/multimet.py
```diff
@@ -583,6 +583,9 @@ def _create_valid_sample_mask(self):
         # Each element is a list of valid integer positions (indexers) for which
         # values are True for a dimension.
         indices = dask.array.nonzero(valid_sample_mask.data)
+        # Compact memory widths. Values are indexes within each mask's shape.
+        dtypes = [np.min_scalar_type(n - 1) for n in valid_sample_mask.shape]
```
omrishefi: Can you use [np.min_scaler_type(n) for...]? Why using n-1? It's simpler to understand if it's n.

omrishefi: Rename dtypes as min_dtypes

omrishefi (commented):
Please see comments

googlehydrology/datasetzoo/multimet.py
amitmarkel: The comment explains: these are indexes. `n` is the total length, there can't be an index with a value of `n`.

amitmarkel: IMHO the fact that it's minimal types is an "impl detail". It's the dtypes for the indices :)

omrishefi: Still using 'n' is more understandable, or even finding the max value, but UTY.

omrishefi: This is a new dtype which represents the smallest possible type that can hold all of the data. It will become more clear if you specify what you do with it. UTY.

amitmarkel: 1. It's the actual max value so self-documenting like `array[len(array) - 1]`.
1. Can't find the true max value, need to compute everything which is a paradox, but there wouldn't be a gain.

omrishefi (dismissed):

googlehydrology/datasetzoo/multimet.py
amitmarkel: Changed to `n` instead (`map`) of `n-1` that's self-documenting, but it's not incorrect though. If it confused you, better to simplify the code thus.

amitmarkel: Done, changed to min_dtypes

omrishefi (approved):

PR #234, opened 2026-02-11
[fix future bug] Set dynamic basin_index vector dtype, not int16.

amitmarkel:
If someone uses more than 2**16 basins with narrow dates, or in the upcoming batchified+lazy mode, this cannot be int16.

I measured np.min_scalar_type performance. It takes 700ns on avg.


omrishefi (dismissed):

amitmarkel:
(merge with main requires re-review)

omrishefi (approved):

PR #238, opened 2026-02-12
Fix multimet dataloader to support dynamic variable product names with underscores

joshsturtevant:
**Summary**
- Fixed bug where `CHIRPS_GEFS` would not parse correctly
- Makes dynamic variable product name parsing in `multimet.py` more robust by normalizing product keys and handling known exceptions (beyond just `ERA5_LAND`)
- Keeps existing config compatibility, but makes the loader more robust

**Key Changes**
- Normalizes product names (case-insensitive, hyphens/underscores handled)
- Adds explicit mapping for products that previously broke parsing (e.g., underscores)
- Handles config keys like `chirps_gefs`, `chirpsgefs`, `CHIRPS_GEFS`, and `CHIRPSGEFS`
- Adds logging when loading Zarr stores for both hindcast and forecast products

**Motivation**
- Previously, the CHIRPS_GEFS product name would not parse correctly due to underscore
- In contrast, the ERA5_LAND product name had a hardwired fix; this PR generalizes the solution

**Testing**
- Ran with configs using various CHIRPS_GEFS and ERA5_LAND naming configurations (w and w/o underscores, etc.)
- Loader correctly resolves to the correct product name and loads data as expected


googlehydrology/datasetzoo/multimet.py
```diff
@@ -1039,32 +1084,44 @@ def _open_zarr(path: Path) -> xr.Dataset:
     return xr.open_zarr(store=path, chunks='auto', decode_timedelta=True)
 
 
-def _get_products_and_bands_from_feature_strings(
-    features: Iterable[str],
-) -> dict[str, list[str]]:
+def _get_products_and_bands_from_feature_dict(feature_dict):
```
grey-nearing: The config input properties are not always dicts. They can be either dicts or lists, depending on whether you want to use a model with vs. without feature groups. This is why the cfg.hindcast_inputs and cfg.forecast_inputs are flattened in lines 182 and 183. 

However, this caused me to notice that the typehints in `~/googlehydrology/utils/config.py` are wrong -- the typehints only show lists, not dicts, and it should be a Union. I'm fixing this presently.

googlehydrology/datasetzoo/multimet.py
```diff
@@ -339,6 +357,15 @@ def __init__(
         LOGGER.debug(f'Dataset on disk: {self._dataset.nbytes / 1024**2} MB')
         LOGGER.debug(f'Sample index size: {sizeof(indices) / 1024**2} MB')
 
+        # TODO(future) :: Move above to the scalar compute block
+        LOGGER.debug('scaler check zero scale')
```
grey-nearing: Could I ask you to pull and merge the most recent changes from `main`? This replication of scaler saving was removed in a recent PR.

googlehydrology/datasetzoo/multimet.py
```diff
@@ -60,6 +60,22 @@
 ]
 MULTIMET_MINIMUM_LEAD_TIME = 1
 
+# Caravan Multimet products that are available in the GCS zarr store
+KNOWN_GCS_PRODUCTS = {
```
grey-nearing: I like the basic strategy here of generalizing the removal of underscores in variable names, but only to known products. Is it necessary to keep a list of known products? It looks like most of these are not used for anything functional, and instead the PRODUCT_ALIASES is doing the heavy lifting. Is there a need to keep this known products list?

googlehydrology/datasetzoo/multimet.py
```diff
@@ -1039,32 +1084,44 @@ def _open_zarr(path: Path) -> xr.Dataset:
...
+    for raw_key, features in feature_dict.items():
+        # Normalize: Uppercase, replace hyphens with underscores
+        norm_key = raw_key.upper().replace("-", "_")
+
+        # Try alias mapping first
+        product = PRODUCT_ALIASES.get(norm_key, norm_key)
+
+        # If it's a known product, use canonical name
+        # (e.g., user says 'era5land', but dataset is 'ERA5_LAND')
+        for known in KNOWN_GCS_PRODUCTS:
```
grey-nearing: Is this loop doing something after the mapping call in line 1111?

grey-nearing (commented):
Hi Josh, 

Thank you SO MUCH for being the first public contributor to GoogleHydrology!!! And thank you for your patience while we worked on the backend to make sure we had a system in place for handling public PRs! 

This is a great fix for CHIRPS. Would you mind taking a look at my questions? It's possible that I've missed something important.

Thanks,
Grey

joshsturtevant:
Thanks Grey - I updated the PR to address your comments (after getting a bit side-tracked for the last few months...)

Changes include:
* Reworked the product/band parsing so it supports both grouped config inputs (`dict[str, list[str]]`) and flat feature lists.
* Removed the separate `KNOWN_GCS_PRODUCTS` list and the follow-on loop over known products; product canonicalization is now handled through a single alias map.
* Merged the latest `main` changes and removed the duplicated scaler-save logic.
* Preserved the previous `union_mapping` fallback behavior so fallback variables are still loaded automatically even when they are not explicitly listed in `hindcast_inputs`.

I also smoke tested the data loader against all available Caravan-Multimet v1.1 products: `CHIRPS`, `CHIRPS_GEFS`, `CPC`, `ERA5_LAND`, `GRAPHCAST`, `HRES`, and `IMERG`. The test run loaded all the training data and entered epoch 1 training successfully.

grey-nearing:
I'm closing this PR because it is now partially superseded by recent code changes. I've migrated the remaining relevant components of this PR to here:

https://github.com/google-research/flood-forecasting/pull/272

PR #244, opened 2026-02-18
Eliminate possible future bug: keep basins order (no `set`)

amitmarkel:
* slightly improve runtime+memory improvement: basins without excluded ones is computed once
* safer self.basins (readonly tuple)

### DIFF
--- a/googlehydrology/evaluation/tester.py
+++ b/googlehydrology/evaluation/tester.py
@@ -109,11 +109,12 @@ class BaseTester(object):
             get_regularization_obj(cfg=self.cfg)
         )
 
-        self._load_run_data()
+        self._load_run_data()  # Sets self.basins
 
         self.dataset = self._get_dataset_all()
 
-        self.exclude_basins = set(self._calc_exclude_basins())
+        exclude_basins = set(self._calc_exclude_basins())
+        self.basins = tuple(e for e in self.basins if e not in exclude_basins)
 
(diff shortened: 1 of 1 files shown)


googlehydrology/evaluation/tester.py
```diff
@@ -109,11 +109,12 @@ def __init__(
             get_regularization_obj(cfg=self.cfg)
         )
 
-        self._load_run_data()
+        self._load_run_data()  # Sets self.basins
 
         self.dataset = self._get_dataset_all()
 
-        self.exclude_basins = set(self._calc_exclude_basins())
+        exclude_basins = set(self._calc_exclude_basins())
```
omrishefi, line 116: Don't we need self.exclude_basins? If it's only local, maybe inline the call in the next line.

omrishefi (commented):
See comment

googlehydrology/evaluation/tester.py
amitmarkel: 1. It was used to setup the local `basins` from `self.basins`, only.
2. Can't inline. We don't want to recompute that for every basin.

omrishefi (approved):

googlehydrology/evaluation/tester.py
amitmarkel: Thanks Omri

PR #246, opened 2026-02-18
[`limit_n_basins` mode] [opt mem] [opt runtime] add mode to allow training and validation on practically any dataset size.

amitmarkel:
`train`:
Using e.g. `limit_n_basins: 100` in the config results in materializing data into memory of only up to 100 basins during training. Data is freed once an epoc is complete.

NOTE: This results in shorter epocs, but users adjust `save_weights_every` and `max_updates_per_epoch` accordingly to their selection.

`validation`:
During validation, support is partial at this time, to simplify this PR to focus on training, since that is the main memory bottleneck and validation isn't usually done over e.g. 46 yrs of data. When enabled (i.e. specified or positive), validation data is loaded only during validation phases, and only up to `validate_n_random_basins` if specified.

`test`/`infer`:
These modes load all data at this time.

---

`limit_n_basins` allows to train on datasets stretching over e.g. 16k basins over 46yrs using ~6GB or so of memory while training.

NOTE: Memory spikes of multimet's init and compute()s are handled in separate.

---

A complementary mode is `lazy_load` which lowers memory usage even further (e.g. 2-3GB) however at significant expense to runtime.

### DIFF
--- a/googlehydrology/datasetzoo/multimet.py
+++ b/googlehydrology/datasetzoo/multimet.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import contextlib
 import functools
 import itertools
 import logging
@@ -322,12 +323,44 @@ class Multimet(Dataset):
         LOGGER.debug('scale data')
         self._dataset = self.scaler.scale(self._dataset)
 
-        if not cfg.lazy_load:
-            LOGGER.debug('[eager load] compute dataset')
+        LOGGER.debug(f'Dataset size: {self._dataset.nbytes / 1024**2} MB')
+        LOGGER.debug('# forecast dataset init complete (%s)', self._period)
+
+        self._dataset_all = self._dataset
+        del self._dataset
+
+    def unload_basins(self) -> None:
+        with contextlib.suppress(AttributeError):
+            del self._dataset
+        with contextlib.suppress(AttributeError):
+            del self._sample_index
+        with contextlib.suppress(AttributeError):
+            del self._num_samples
+        with contextlib.suppress(AttributeError):
+            del self._per_basin_target_stds
+
+    def load_basins(self, basins: list[str] | None = None) -> None:
+        self._data_cache: dict[str, xr.DataArray] = {}
+        self.unload_basins()
+
+        if basins is None:
+            self._dataset = self._dataset_all
+        else:
+            LOGGER.debug('[limit %d basins] (%s)', len(basins), self._period)
+            self._dataset = self._dataset_all.sel(basin=basins)
+
+        if not self._cfg.lazy_load:
+            LOGGER.debug('[eager load] compute dataset (%s)', self._period)
             (self._dataset,) = dask.compute(self._dataset)
             memory.release()
         else:
-            LOGGER.debug('[lazy load] not computing dataset')
+            LOGGER.debug('[lazy load] not computing dataset (%s)', self._period)
+
+        LOGGER.debug(
+            'Dataset size: %f MB (%s)',
+            self._dataset.nbytes / 1024**2,
+            self._period,
+        )
 
         LOGGER.debug('create valid sample mask and indices plan')
         valid_sample_mask, indices = self._create_valid_sample_mask()
@@ -335,9 +368,11 @@ class Multimet(Dataset):
         (indices,) = dask.compute(indices)
         memory.release()
 
-        LOGGER.debug(f'Dataset size: {sizeof(self._dataset) / 1024**2} MB')
-        LOGGER.debug(f'Dataset on disk: {self._dataset.nbytes / 1024**2} MB')
-        LOGGER.debug(f'Sample index size: {sizeof(indices) / 1024**2} MB')
+        LOGGER.debug(
+            'Sample index size: %f MB (%s)',
+            sizeof(indices) / 1024**2,
+            self._period,
+        )
 
         # Create sample index lookup table for `__getitem__`.
         LOGGER.debug('create sample index')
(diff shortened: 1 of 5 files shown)


omrishefi (dismissed):

googlehydrology/datasetzoo/multimet.py
```diff
@@ -322,22 +323,56 @@ def __init__(
...
+
+    def unload_basins(self) -> None:
+        with contextlib.suppress(AttributeError):
+            del self._dataset
+        with contextlib.suppress(AttributeError):
+            del self._sample_index
+        with contextlib.suppress(AttributeError):
+            del self._num_samples
+        with contextlib.suppress(AttributeError):
+            del self._per_basin_target_stds
```
omrishefi, line 340: Can you merge all lines to use a single "with" clause?

googlehydrology/datasetzoo/multimet.py
```diff
@@ -322,22 +323,56 @@ def __init__(
         LOGGER.debug('scale data')
         self._dataset = self.scaler.scale(self._dataset)
 
-        if not cfg.lazy_load:
-            LOGGER.debug('[eager load] compute dataset')
+        LOGGER.debug(f'Dataset size: {self._dataset.nbytes / 1024**2} MB')
+        LOGGER.debug('# forecast dataset init complete (%s)', self._period)
+
+        self._dataset_all = self._dataset
+        del self._dataset
```
omrishefi, line 330: Why del here?

googlehydrology/datasetzoo/multimet.py
```diff
@@ -322,22 +323,56 @@ def __init__(
...
             (self._dataset,) = dask.compute(self._dataset)
             memory.release()
         else:
-            LOGGER.debug('[lazy load] not computing dataset')
+            LOGGER.debug('[lazy load] not computing dataset (%s)', self._period)
+
+        LOGGER.debug(
+            'Dataset size: %f MB (%s)',
+            self._dataset.nbytes / 1024**2,
+            self._period,
```
omrishefi, line 362: Why printing self._period in all debug lines?

omrishefi (commented):
Let's review the changes together

googlehydrology/datasetzoo/multimet.py
amitmarkel: Want it to be safe, otherwise a line failing means `del`s stop at that line.

I considered a loop on value names and `__dict__.pop(name, None)` but something like that is harder to read and doesn't link symbols and limits search. This way, `self.<field name>` is findable.

googlehydrology/datasetzoo/multimet.py
amitmarkel: Want `self._dataset` not be defined for safety and clarity below. Another option is to rename all the way up dataset to dataset_all or assess if can just keep without `self.` until the end, but that would pollute this PR with refactor, could add a todo.

googlehydrology/datasetzoo/multimet.py
amitmarkel: Helped see it faster as a standalone line (contextless). Can remove though.

googlehydrology/datasetzoo/multimet.py
```diff
@@ -322,22 +323,56 @@ def __init__(
...
+            del self._dataset
+        with contextlib.suppress(AttributeError):
+            del self._sample_index
+        with contextlib.suppress(AttributeError):
+            del self._num_samples
+        with contextlib.suppress(AttributeError):
+            del self._per_basin_target_stds
+
+    def load_basins(self, basins: list[str] | None = None) -> None:
+        self._data_cache: dict[str, xr.DataArray] = {}
```
omrishefi, line 343: Move to unload_basins

googlehydrology/datasetzoo/multimet.py
```diff
@@ -322,22 +323,56 @@ def __init__(
...
+        with contextlib.suppress(AttributeError):
+            del self._dataset
+        with contextlib.suppress(AttributeError):
+            del self._sample_index
+        with contextlib.suppress(AttributeError):
+            del self._num_samples
+        with contextlib.suppress(AttributeError):
+            del self._per_basin_target_stds
+
+    def load_basins(self, basins: list[str] | None = None) -> None:
```
omrishefi, line 342: Add a comment

googlehydrology/evaluation/tester.py
```diff
@@ -112,9 +112,9 @@ def __init__(
         self._load_run_data()  # Sets self.basins
 
         self.dataset = self._get_dataset_all()
-
-        exclude_basins = set(self._calc_exclude_basins())  # Needs self.dataset
-        self.basins = [e for e in self.basins if e not in exclude_basins]
+        if cfg.limit_n_basins < 1:
+            self.dataset.load_basins()
+            self.basins = self._calc_and_apply_excluded_basins(self.basins)
```
omrishefi, line 117: It seems like a change of logic. Let's not pass self.basins to the function.

googlehydrology/training/basetrainer.py
```diff
@@ -179,6 +185,54 @@ def _freeze_model_parts(self):
                 f'Could not resolve the following module parts for finetuning: {unresolved_modules}'
             )
 
+    def init_loader(
```
omrishefi, line 188: Please split into smaller PRs

omrishefi (commented):
Please see comments

sylvesterkaczmarek (commented):
`limit_n_basins` is described as an upper bound, but when it exceeds the number of available basins the wrapped `np.take(..., mode='wrap')` repeats basin IDs until the requested count is reached. That duplicates training samples and changes their weighting instead of simply loading all basins once. Could `max_random_basins` be clamped to `len(self.basins)` before constructing the window?

PR #265, opened 2026-08-11
Fix CSV hindcast schema loading

akshusuba:
## Summary
- load `union_mapping` source variables alongside requested CSV hindcast features
- assign lead day 1 to forecast-origin historical inputs that are also configured forecast features
- retain ordinary date-only dimensions for nowcasts, reanalysis, and union-source variables
- verify union completion, forecast minimum lead, and exact historical sample alignment

This is intentionally separate from #264. The implementation now matches the control-basin compatibility shim: HRES/GraphCast-style historical inputs retain lead-day semantics, while CPC/IMERG/ERA5-Land-style inputs do not.

## Test plan
- [x] `/tmp/googlehydrology-venv/bin/pytest -q test/test_multimet.py -k 'csv_hindcast'` (`2 passed`)
- [x] `git diff --check`
- [ ] `/tmp/googlehydrology-venv/bin/pytest -q test/test_multimet.py` (previously `13 passed`, with one existing `test_forecast_dataset_init_success` scaler-call-order failure also present on `main`)

### DIFF
--- a/googlehydrology/datasetzoo/multimet.py
+++ b/googlehydrology/datasetzoo/multimet.py
@@ -738,12 +738,30 @@ class Multimet(Dataset):
                 ]
             )
         )
-        return load_caravan_timeseries_together(
+        dataset = load_caravan_timeseries_together(
             data_dir=self._dynamics_data_path,
             basins=self._basins,
             target_features=features,
             csv=True,
         )
+        forecast_origin_features = [
+            feature
+            for feature in self._hindcast_features
+            if feature in self._forecast_features
+        ]
+        plain_features = [
+            feature
+            for feature in features
+            if feature not in forecast_origin_features
+        ]
+        datasets = [dataset[plain_features]]
+        if forecast_origin_features:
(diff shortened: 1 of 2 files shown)


akshusuba:
Follow-up validation after correcting the historical lead semantics:

- The CSV loader now assigns lead day 1 only to hindcast variables that are also configured forecast inputs; CPC/IMERG/ERA5-Land-style variables and union sources remain date-only.
- The focused test checks exact extracted tensors and verifies that forecast issue dates plus lead day 1 align to the same valid dates as ordinary historical variables.
- The control experiment was rerun with a stricter comparator that first requires identical NaN and finite-value masks, then compares only finite positions.

Strict control result on `camels_12377150`:
- canonical samples: 2
- CSV samples: 2
- sample max absolute difference: `2.3841858e-07`
- model output max absolute difference: `1.4305115e-06`
- tolerance: `1e-5`
- status: validated

grey-nearing:
Thank you for this Pull Request! We are changing the input file formats for this codebase and no longer using CSV files. I am closing this PR without merging based on these updates.

PR #266, opened 2026-08-11
Reject unusable CSV forecast fallback

akshusuba:
## Summary
- require each CSV forecast feature to use its dedicated issue-date by lead-time file
- reject the basin-wide historical CSV as a forecast fallback with a clear schema error
- retain and test the valid dedicated forecast-file path

This is intentionally independent from #264 and #265. The removed fallback assigned lead zero and then sliced to supported leads 1–N, returning an empty forecast dimension that could not produce a valid sample.

## Test plan
- [x] `/tmp/googlehydrology-venv/bin/pytest -q test/test_multimet.py -k 'csv_forecast'` (`2 passed`)
- [x] `git diff --check`

### DIFF
--- a/googlehydrology/datasetzoo/multimet.py
+++ b/googlehydrology/datasetzoo/multimet.py
@@ -746,13 +746,10 @@ class Multimet(Dataset):
         return ds
 
     def _load_forecast_as_csv(self) -> xr.Dataset:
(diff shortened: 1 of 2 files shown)


grey-nearing:
Thank you for this Pull Request! We are changing the input file formats for this codebase and no longer using CSV files. I am closing this PR without merging based on these updates.

PR #268, opened 2026-08-11
Feature: Add 'Hot Start' Inference Support to LSTM Architectures

grey-nearing:
## Description
This PR introduces a "hot start" feature that allows users to save the final cell and hidden states of the hindcast LSTMs and subsequently resume inference initialized directly from those states. 
This skips the memory and computational overhead of re-simulating the entire `seq_length` hindcast spin-up period and removes the requirement to load the full historical timeseries datasets when generating day-of forecasts.
### Key Capabilities and Changes
- **No-Pickle State Persistence:** Model states are structurally exported and imported via Numpy's `np.savez_compressed` (avoiding standard `torch.save()` pickle wrappers) to guarantee cross-machine portability and security. Both hidden states (`h`) and cell states (`c`) are preserved for hindcast and forecast networks.
- **`HandoffForecastLSTM` Logic:** Rebuilt the sequence forward pass to conditionally bypass the `handoff_network` and overlap merging during hot start mode (`seq_length = 0`), directly initializing hidden and cell states for the `forecast_lstm`.
- **`MeanEmbeddingForecastLSTM` Logic:** Updated forward pass and embedding masking to accept initial state injections sequentially on hindcast and forecast LSTM blocks.
- **Evaluation Runner Integration (`tester.py`):**
  - **Per-Basin Pre-Loading:** Pre-loads state archives once per basin in `BaseTester._evaluate` from either a single file or a directory (`state_{basin}.npz` / `{basin}.npz`), eliminating per-step disk I/O.
  - **Automatic State Saving:** Automatically persists final states to `<run_dir>/hot_start_states/state_{basin}.npz` during evaluation when `save_state: True`.
  - **Batch Size Validation:** Enforces `batch_size = 1` for hot-start runs with descriptive error messaging.
- **Configuration Support (`config.py`):** Added `save_state`, `hot_start_path`, and `seq_length` properties and setters to `Config`.
- **Validation Fallbacks:** Added `seq_length <= 0` checks in `validate_samples.py` so multidimensional datasets process zero-sized historical sequences gracefully.
- **Hot Start Configuration Template:** Added `example-configs/template_hot_start_config.yml` as boilerplate for hot-start execution.
## Testing
- Added `test/test_hot_start.py` with 4 automated, self-contained unit tests using pytest's `tmp_path` fixture (fully compatible with GitHub Actions CI):
  1. `test_handoff_forecast_lstm_hot_start`: Verified cold vs. hot start numerical parity ($< 10^{-5}$ max absolute diff).
  2. `test_mean_embedding_forecast_lstm_hot_start`: Verified cold vs. hot start numerical parity ($< 10^{-5}$ max absolute diff).
  3. `test_hot_start_preloading`: Verified in-memory state restoration via `load_state_from_disk`.
  4. `test_hot_start_batch_size_validation`: Verified `batch_size != 1` validation in `RegressionTester`.

### DIFF
--- a/googlehydrology/modelzoo/basemodel.py
+++ b/googlehydrology/modelzoo/basemodel.py
@@ -108,6 +108,8 @@ class BaseModel(nn.Module):
     ) -> None:
         """Save the hot start state of the model.
 
+        Default implementation is a no-op for models that do not support state persistence.
+
         Parameters
         ----------
         data : dict[str, torch.Tensor | dict[str, torch.Tensor]]
@@ -115,17 +117,23 @@ class BaseModel(nn.Module):
         path : str | Path
             The file path where the state should be saved (npz recommended).
         """
-        raise NotImplementedError
+        pass
 
     def load_state_from_disk(self, path: str | Path) -> None:
         """Pre-load a hot start state archive from disk into memory.
 
+        Default implementation is a no-op for models that do not support state persistence.
+
         Parameters
         ----------
         path : str | Path
-            Path to the .npz state file to load.
+            Path to the state file to load.
         """
-        self._preloaded_state = dict(np.load(path, allow_pickle=False))
+        pass
+
+    def reset_state(self) -> None:
+        """Reset in-memory hot-start state between basins."""
+        self._preloaded_state = None
 
     def pre_model_hook(
         self, data: dict[str, torch.Tensor], is_train: bool
--- a/googlehydrology/modelzoo/basemodel.py
+++ b/googlehydrology/modelzoo/basemodel.py
@@ -13,6 +13,9 @@
 # limitations under the License.
 
 
+from pathlib import Path
+
+import numpy as np
 import torch
 import torch.nn as nn
 
@@ -47,6 +50,7 @@ class BaseModel(nn.Module):
             scaler_dir=(cfg.base_run_dir if cfg.is_finetuning else cfg.run_dir),
             calculate_scaler=False,
         )
+        self._preloaded_state = None
 
     def sample(
         self,
@@ -96,6 +100,32 @@ class BaseModel(nn.Module):
             Model output and potentially any intermediate states and activations as a dictionary.
         """
         raise NotImplementedError
+
+    def save_state(
+        self,
+        data: dict[str, torch.Tensor | dict[str, torch.Tensor]],
+        path: str | Path,
+    ) -> None:
+        """Save the hot start state of the model.
+
+        Parameters
+        ----------
+        data : dict[str, torch.Tensor | dict[str, torch.Tensor]]
+            Dictionary containing input features as key-value pairs.
+        path : str | Path
+            The file path where the state should be saved (npz recommended).
+        """
+        raise NotImplementedError
+
+    def load_state_from_disk(self, path: str | Path) -> None:
+        """Pre-load a hot start state archive from disk into memory.
+
+        Parameters
+        ----------
+        path : str | Path
+            Path to the .npz state file to load.
+        """
+        self._preloaded_state = dict(np.load(path, allow_pickle=False))
 
     def pre_model_hook(
         self, data: dict[str, torch.Tensor], is_train: bool
--- a/googlehydrology/modelzoo/handoff_forecast_lstm.py
+++ b/googlehydrology/modelzoo/handoff_forecast_lstm.py
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from pathlib import Path
+
+import numpy as np
 import torch
 import torch.nn as nn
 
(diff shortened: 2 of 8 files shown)


googlehydrology/modelzoo/handoff_forecast_lstm.py
```diff
@@ -305,7 +424,8 @@ def forward(self, data: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
                     y_hindcast_overlap[key], 
                     y_forecast[key][:, -self.lead_time :, :]
                 ], dim=1
-            )[:, -self.seq_length :, :] for key in y_spinup
```
ashermetzger-rgb, line 308: Removing `[:, -self.seq_length :, :]` breaks cold start and I think it doesn't affect hot start.
Cold-start (seq_length=365): [:, -365:, :] → trims to seq_length
Hot-start (seq_length=0): [:, -0:, :] → returns full tensor

googlehydrology/modelzoo/handoff_forecast_lstm.py
```diff
@@ -274,24 +277,140 @@ def forward(self, data: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
...
-        overlap_embeddings = hindcast_embeddings[:, -self.overlap :,]
-        spinup, (h_hindcast, c_hindcast) = self.hindcast_lstm(spinup_embeddings)
-        hindcast_overlap, _ = self.hindcast_lstm(
-            overlap_embeddings, (h_hindcast, c_hindcast)
-        )
+        state = getattr(self, '_preloaded_state', None)
+        if state is None:
+            hot_start_path = getattr(self.cfg, 'hot_start_path', None)
+            if hot_start_path is not None and Path(hot_start_path).is_file():
+                state = dict(np.load(hot_start_path, allow_pickle=False))
```
ashermetzger-rgb: Remove.
The Tester already calls model.load_state_from_disk(path).
And it is both slow and breaks torch.compile.

googlehydrology/modelzoo/basemodel.py
```diff
@@ -97,6 +101,32 @@ def forward(
...
+
+    def load_state_from_disk(self, path: str | Path) -> None:
+        """Pre-load a hot start state archive from disk into memory.
+
+        Parameters
+        ----------
+        path : str | Path
+            Path to the .npz state file to load.
+        """
+        self._preloaded_state = dict(np.load(path, allow_pickle=False))
```
ashermetzger-rgb: Maybe nit:
The assumption is that every model will always save state as .npz. That's true today (only two LSTM models exist), but BaseModel is supposed to be a generic base for any model architecture.

The fix suggestion is simple: make load_state_from_disk also raise NotImplementedError, and move the np.load implementation into HandoffForecastLSTM and MeanEmbeddingForecastLSTM where the concrete .npz save logic already lives.

test/test_hot_start.py
ashermetzger-rgb, line 1: Some tests are missing.
The gaps are scenarios that could silently break without dedicated tests:
  
1. Multi-basin state isolation: The Tester loops over basins and calls load_state_from_disk() per basin. If _preloaded_state isn't properly reset  
between basins, basin B's forecast could use basin A's river states. No test currently evaluates two basins sequentially.
2. Cold-start shape regression: Given Comment 1 (the output shape bug), there should be an explicit assertion that y_hat.shape[1] == seq_length    
still holds for cold-start runs after this PR. Otherwise the shape change goes undetected.
3. Corrupt/incompatible state files: What happens if the .npz file has missing keys (e.g., no h_forecast), wrong hidden dimensions, or was saved by
a different model architecture? Currently no test covers this — the code would crash with a confusing KeyError or silent shape mismatch.           
4. forecast_overlap > 0 equivalence: The hot-start path has special handling for when forecast overlap is active (line 369: forecast_embeddings[:, 
-self.lead_time:, :]). No test exercises this branch.

I can take care of this later if you want.

ashermetzger-rgb (changes requested):
Hey Grey,
A few comments.
Let me know how you want us to handle them.

googlehydrology/modelzoo/handoff_forecast_lstm.py
grey-nearing: Fixed.

googlehydrology/modelzoo/handoff_forecast_lstm.py
grey-nearing: Fixed.

googlehydrology/modelzoo/basemodel.py
grey-nearing: Very good point. Added a passthrough method to the base class and put the working methods in the child classes.

test/test_hot_start.py
grey-nearing: Added.

ashermetzger-rgb (approved):
Thanks!

PR #271, opened 2026-08-12
Unify input file formats to Zarr and generalize dataloaders

grey-nearing:
### Summary of Changes
This PR standardizes input file storage and data loaders across the codebase onto the **Zarr format**, improves data loader flexibility, and provides seamless migration tooling:
1. **Unified Zarr Input Format & Generalized Dataloaders:**
   * Standardized Caravan static attributes (`attributes.zarr`) and streamflow time series (`streamflow.zarr`) into consolidated, chunked Zarr stores.
   * Generalized the `Multimet` dataloader with robust path resolution—supporting single consolidated stores (`timeseries.zarr`), per-product subdirectories (`<product>/timeseries.zarr`), and case-insensitive store matching.
   * Removed hardcoded basin naming assumptions (such as `basin.split('_')[0]`), allowing arbitrary basin identifier naming conventions.
2. **Feature Scaler Migration (`scaler.zarr`):**
   * Migrated feature normalization scalers to Zarr format (`scaler.zarr`) while retaining backward-compatible fallback to existing `scaler.nc` files.
3. **Caravan Dataset Conversion Utilities:**
   * Introduced `googlehydrology.datautils.convert` (`convert_caravan_attributes`, `convert_caravan_timeseries`, `convert_caravan_to_zarr`) for converting legacy NetCDF and CSV directories into consolidated Zarr stores.
   * Exposed the conversion tool via the CLI entrypoint: `run convert-caravan`.
4. **Output Zarr Store Metadata Consolidation:**
   * Updated `BaseTester.evaluate()` to automatically write consolidated metadata for `{period}_results.zarr` at the completion of inference.
5. **CMAL Sampling & Negative Sample Handling:**
   * Extended `sample_cmal_deterministic` to accept `scaler` and apply analytical quantile zero-bounding through `_handle_negative_values`.
---
### Verification & Testing
* **Numerical Parity:** Validated identical inference results between legacy NetCDF inputs and converted Zarr stores on both the 5-basin tutorial model and the full 615-catchment FloodHub operational benchmark.
* **Test Suite:** All 212 unit and end-to-end integration tests pass successfully (`pytest test/`).


googlehydrology/datasetzoo/multimet.py
```diff
@@ -688,16 +689,33 @@ def _load_hindcast_features(self) -> list[xr.Dataset]:
...
+
+        Returns
+        -------
+        list[xr.Dataset]
+            Datasets containing the loaded hindcast features.
+        """
         # Prepare hindcast features to load, including the masks of union_mapping
         features = set(self._hindcast_features) | set(
             (self._union_mapping or {}).values()
         )
```
grey-nearing: **Nit (determinism):** iterating a `set` of strings gives a different
`ds.data_vars` order on every process, since CPython randomizes string hashes.

As noted in the main review this does *not* corrupt model input order —
`_extract_hindcasts` re-derives order from `self._hindcast_features`, which is
config-ordered — so this is about reproducible artifacts and readable debug
output, not correctness.

`dict.fromkeys` keeps the dedup you wanted from `set` while making the order
deterministic:

```suggestion
        features = list(
            dict.fromkeys(
                [*self._hindcast_features, *(self._union_mapping or {}).values()]
            )
        )
```

grey-nearing: **Please build this list with `dict.fromkeys` instead of `set`, so the feature order is the same on every run.**

Iterating a `set` of strings gives a different `ds.data_vars` order in each
process, since CPython randomizes string hashes. Model input order is unaffected
(`_extract_hindcasts` re-derives it from the config-ordered
`self._hindcast_features`), so this is about reproducible artifacts and readable
debug output rather than correctness.

`dict.fromkeys` keeps the dedup you wanted from `set` while making the order
deterministic:

```suggestion
        features = list(
            dict.fromkeys(
                [*self._hindcast_features, *(self._union_mapping or {}).values()]
            )
        )
```

grey-nearing (commented):
**The nits from this review are fixed directly in 283e161, so nothing is
outstanding for the branch.** Recording what changed and what is left to decide.

### Fixed on this branch

- `_load_hindcast_as_zarr` built its feature list from a `set`, so `ds.data_vars`
  came out in a different order in every process. Model input order was never
  affected — `_extract_hindcasts` re-derives it from the config-ordered
  `self._hindcast_features` — so this was reproducibility and debuggability
  rather than correctness. It now uses `dict.fromkeys`, which dedups while
  preserving insertion order. `multimet.py` has no other `set()` call sites.
- `NUMERICAL_EQUIVALENCE_VERIFICATION_PLAN.md` and `UNIFY_INPUT_FORMATS_PLAN.md`
  removed from the repo root. Both were working documents for this PR,
  introduced by its own commit `de93d39`, and referenced nowhere in the tree.

### Numerical equivalence, now measured

The verification plan was never executed: there is no
`verify_numerical_equivalence.py` and no report anywhere on the branch. The
comparison is available directly though, because both runs are committed at the
same epoch with a byte-identical `model_epoch015.pt` (same md5).

| | |
|---|---|
| cells compared | 112 (14 metrics x 8 basins) |
| bit-identical | 63 |
| max absolute difference | `6.1e-4` (FLV, `camels_13235000`) |
| max relative difference | `6.1e-6` (Beta-NSE, `camels_04216418`) |

Every one of the 49 differing cells sits within ~100x float32 epsilon in relative
terms, which is what reordered float32 accumulation looks like rather than a
change in what is being computed. So the Zarr path is numerically equivalent to
the legacy path at float32 precision.

This does **not** satisfy the deleted plan's stated criterion of `|delta| < 1e-5`,
but that criterion was mis-specified: an absolute tolerance is not meaningful for
a metric such as FLV whose values reach -1606. The relative figure is the one to
read.

Full suite green after the changes: 220 passed.

### Left to decide

Merge order against #272. `pr/271` is not an ancestor of `pr/272`; they are
independent branches that both rewrite `multimet.py` heavily, so whichever lands
second needs real conflict resolution rather than a fast-forward.

### Coverage caveat

At 483 changed files, line-by-line coverage by a single reviewer is not a credible
claim. This review covers the ordering question, the scaler backward-compatibility
path, merge mechanics, and the equivalence check above.

PR #283, opened 2026-09-02
kruparell-data assimilation

kruparell:
# Variational Data Assimilation for Hydrological Forecasting

## Summary
This PR introduces a gradient-based Variational Data Assimilation (DA) framework for hydrological forecasting. This framework optimizes latent model components (such as static and dynamic catchment representations) exposed through a standardized, model-agnostic interface.

### 1. `googlehydrology/modelzoo/basemodel.py` & `mean_embedding_forecast_lstm.py`
* **`BaseModel` contract**: Added `supported_assimilation_components` property and `get_supported_assimilation_components()` helper.
* **`MeanEmbeddingForecastLSTM` overrides**: 
  * Checks `data['assimilation_overrides']` (with backward-compatible top-level keys) for `'static_embedding'`, `'hindcast_embedding'`, and `'forecast_embedding'`.
  * If supplied, reuses the optimized representations directly instead of re-evaluating embedding submodules.
  * Attaches all generated representations to the returned dictionary for downstream inspection.

### 2. `googlehydrology/utils/cmal_deterministic.py`
* Standalone mathematical utilities for CMAL:
  * `calc_cmal_mean(mu, b, tau)`: Vectorized numerical quadrature computing the expected value $\mathbb{E}[y]$ from asymmetric Laplace parameters.
  * `ensure_y_hat(pred, use_median=True/False)`: Unifies prediction dictionaries and ensures `y_hat` is populated with robust handling of non-finite outputs.

### 3. `googlehydrology/utils/assimilationconfig.py`
* Config dataclass validating DA hyperparameters: `assimilation_window`, `assimilation_lead_time`, `assimilation_components`, `regularization_weight`, `learning_rate`, `epochs`, and component-specific learning rates and regularization weights.

### 4. `googlehydrology/evaluation/assimilation.py`
* Self-contained Var DA assimilation engine:
  * Optimizes exposed latent representations over rolling assimilation windows using observed streamflow.
  * Applies intensive scale-invariant regularization (`torch.mean((param - base_param) ** 2)`).
  * Implements local `_create_assimilation_optimizer` supporting Adam, AdamW, SGD, RMSprop, and Adagrad without altering package-level training contracts.
  * Updates evaluation metrics and rollout predictions across sequential time horizons.

### 5. `googlehydrology/run.py`, `evaluate.py`, & `tester.py`
* Integrates `run_data_assimilation` flag into CLI commands and pipeline workflows (`evaluate` and `infer`).
* Updates documentation in `README.md` and `docs/source/usage/quickstart.rst`.

### 6. `test/test_assimilation.py` & `test/test_cli.py`
* Comprehensive test suite validating:
  * Multi-component embedding optimization with model weight freezing.
  * Model-agnostic fallback on mock architectures implementing the contract.
  * Regularization penalty enforcement and loss minimization.
  * CLI and pipeline entrypoints with Data Assimilation enabled.

---

## Verification & Testing

All unit and integration tests pass cleanly:

```bash
$ pytest test/test_assimilation.py test/test_cli.py -v
============================= test session starts ==============================
test/test_assimilation.py::AssimilationTest::test_initialization PASSED
test/test_assimilation.py::AssimilationTest::test_embedding_gradient_update PASSED
test/test_assimilation.py::AssimilationTest::test_model_weights_remain_frozen PASSED
test/test_assimilation.py::AssimilationTest::test_regularization_loss PASSED
test/test_assimilation.py::AssimilationTest::test_component_specific_learning_rates PASSED
test/test_assimilation.py::AssimilationTest::test_missing_component_handling PASSED
test/test_assimilation.py::AssimilationTest::test_generic_model_contract PASSED
test/test_assimilation.py::AssimilationTest::test_rollout_prediction_stitching PASSED
...
test/test_cli.py::test_run_main_infer_data_assimilation PASSED
======================== 19 passed in 23.96s ========================
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,12 +110,24 @@ Calculate performance metrics (NSE, KGE) on the test set:
    run evaluate --run-dir /path/to/your/model_run/
    ```
 
+To run evaluation with test-time 4D-Var Data Assimilation:
+
+   ```
+   run evaluate --run-dir /path/to/your/model_run/ --data-assimilation
+   ```
+
 ### **Inference**
 
 Generate predictions (without skipping NaN observations):
    
    ```
    run infer --run-dir /path/to/your/model_run/
+   ```
+
+To run inference with test-time 4D-Var Data Assimilation (assimilating available hindcast observations before generating forecasts):
+
+   ```
+   run infer --run-dir /path/to/your/model_run/ --data-assimilation
    ```
 
 ## **Configuration**
--- a/googlehydrology/evaluation/assimilation.py
+++ b/googlehydrology/evaluation/assimilation.py
@@ -1,0 +1,548 @@
+"""Model-agnostic Data Assimilation (DA) engine for hydrological forecasting models."""
+
+import logging
+import re
+from typing import Any, Dict, List, Optional, Tuple, Union
+import warnings
+
+import numpy as np
+import torch
+import torch.nn as nn
+
+from googlehydrology.evaluation.metrics import calculate_metrics, get_available_metrics
+from googlehydrology.modelzoo.basemodel import BaseModel
+from googlehydrology.training import get_loss_obj, get_regularization_obj
+from googlehydrology.utils.assimilationconfig import AssimilationConfig
+from googlehydrology.utils.cmal_deterministic import calc_cmal_mean, ensure_y_hat
+
+logger = logging.getLogger(__name__)
+
+# Backward-compatibility alias
+_ensure_y_hat = ensure_y_hat
+
+
+def _create_assimilation_optimizer(
+    param_groups: List[Dict[str, Any]], cfg: Any
+) -> torch.optim.Optimizer:
+    """Create optimizer for data assimilation parameter groups without modifying model training contracts."""
+    optimizer_name = getattr(cfg, 'optimizer', 'adam').lower()
+    default_lr = getattr(cfg, 'initial_learning_rate', getattr(cfg, 'learning_rate', 0.01))
+    if isinstance(default_lr, dict):
+        default_lr = default_lr.get(0, 0.01)
+
+    if optimizer_name == 'adam':
+        return torch.optim.Adam(param_groups, lr=default_lr)
+    elif optimizer_name == 'adamw':
+        return torch.optim.AdamW(param_groups, lr=default_lr)
+    elif optimizer_name == 'sgd':
+        return torch.optim.SGD(param_groups, lr=default_lr)
+    elif optimizer_name == 'asgd':
+        return torch.optim.ASGD(param_groups, lr=default_lr)
+    elif optimizer_name == 'rmsprop':
+        return torch.optim.RMSprop(param_groups, lr=default_lr)
+    elif optimizer_name == 'adagrad':
+        return torch.optim.Adagrad(param_groups, lr=default_lr)
+    elif optimizer_name == 'adadelta':
+        return torch.optim.Adadelta(param_groups, lr=default_lr)
+    elif optimizer_name == 'adamax':
+        return torch.optim.Adamax(param_groups, lr=default_lr)
+    else:
+        raise NotImplementedError(
+            f'{optimizer_name} not implemented or not supported for Data Assimilation'
+        )
+
+
+def _copy_data_dict(data_dict: dict) -> dict:
+    result = {}
+    for key, val in data_dict.items():
+        if isinstance(val, dict):
+            result[key] = _copy_data_dict(val)
+        elif isinstance(val, torch.Tensor):
+            result[key] = val.clone()
+        elif isinstance(val, np.ndarray):
+            result[key] = val.copy()
+        else:
+            result[key] = val
+    return result
+
+
+def _infer_sequence_length(batch_dict: dict) -> Optional[int]:
+    """Helper to infer sequence length from target y or 3D feature tensors."""
+    if 'y' in batch_dict and isinstance(batch_dict['y'], (torch.Tensor, np.ndarray)) and batch_dict['y'].ndim >= 2:
+        return batch_dict['y'].shape[1]
+    for val in batch_dict.values():
+        if isinstance(val, (torch.Tensor, np.ndarray)) and val.ndim == 3:
+            return val.shape[1]
+        elif isinstance(val, dict):
+            sub_sequence_length = _infer_sequence_length(val)
+            if sub_sequence_length is not None:
+                return sub_sequence_length
+    return None
+
+
+def _get_variable_learning_rate(learning_rate_config: Any, variable_name: str) -> float:
+    """Retrieves target-specific learning rate from learning_rate_config."""
+    if isinstance(learning_rate_config, dict):
+        if variable_name in learning_rate_config:
+            return float(learning_rate_config[variable_name])
+        return float(list(learning_rate_config.values())[0])
+    elif isinstance(learning_rate_config, (list, tuple)):
+        return float(learning_rate_config[0])
+    else:
+        return float(learning_rate_config)
+
+
+# Backward-compatibility alias
+_get_var_lr = _get_variable_learning_rate
+
+
+def _slice_hydrology_batch(batch_dict: dict, slice_start_step: int, slice_end_step: int) -> dict:
+    """Fast, zero-copy slicing of 3D sequence tensors in a googlehydrology batch dict."""
+    non_sequence_keys = {
+        'x_s', 'x_one_hot', 'static_features',
+        'last_prediction', 'assimilation_overrides',
+    }
+    hindcast_length, forecast_length = None, None
+    if 'x_d_hindcast' in batch_dict and isinstance(batch_dict['x_d_hindcast'], dict):
+        for val in batch_dict['x_d_hindcast'].values():
+            if isinstance(val, torch.Tensor) and val.ndim == 3:
+                hindcast_length = val.shape[1]
+                break
+    if 'x_d_forecast' in batch_dict and isinstance(batch_dict['x_d_forecast'], dict):
+        for val in batch_dict['x_d_forecast'].values():
+            if isinstance(val, torch.Tensor) and val.ndim == 3:
+                forecast_length = val.shape[1]
+                break
+    lead_delta = (
+        (forecast_length - hindcast_length)
+        if (hindcast_length is not None and forecast_length is not None and forecast_length > hindcast_length)
+        else 0
+    )
+
+    result = {}
+    for key, val in batch_dict.items():
+        if isinstance(val, dict) and key != 'assimilation_overrides':
+            if key in ('x_d_forecast', 'forecast_features'):
+                result[key] = _slice_hydrology_batch(val, slice_start_step, slice_end_step + lead_delta)
+            else:
+                result[key] = _slice_hydrology_batch(val, slice_start_step, slice_end_step)
+        elif isinstance(val, torch.Tensor) and key not in non_sequence_keys and val.ndim == 3:
+            sequence_length = val.shape[1]
+            slice_end_index = (
+                min(slice_end_step + lead_delta, sequence_length)
+                if key in ('x_d_forecast', 'y')
+                else min(slice_end_step, sequence_length)
+            )
+            result[key] = val[:, min(slice_start_step, sequence_length):slice_end_index, :]
+        elif isinstance(val, np.ndarray) and key in ('date', 'y') and val.ndim == 2:
+            sequence_length = val.shape[1]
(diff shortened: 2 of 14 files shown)


googlehydrology/evaluation/assimilation.py
```diff
@@ -0,0 +1,666 @@
...
+                                chunk_data['last_prediction'] = last_prediction_curr
+
+                            pred_dict = ensure_y_hat(model(chunk_data), use_median=False)
+                            p_sub = pred_dict['y_hat'][:, :win_len, :]
+                            if p_sub.ndim == 2: p_sub = p_sub.unsqueeze(-1)
+                            t_sub = chunk_data['y'][:, :win_len, :]
+
+                            mask = ~torch.isnan(t_sub) & ~torch.isnan(p_sub)
+                            if mask.any():
+                                loss = torch.mean((p_sub[mask] - t_sub[mask]) ** 2)
```
grey-nearing: Perhaps we should add the loss function to training/loss.py as a child of baseloss. And then either: (1) add an assimilation_loss arguemnt to config.py, or (2) have a specific base_loss inheritance class that is an assimilation_loss and hard-code assimilation.py to use that. The former is better unless you want to make sure the user is unable to change the assimilation loss. I would do that only if you think there is a potential for a user to make a mistake in the sense that assimilation can theoretically or practically only use one specific loss. If there are specific loss functions you want to allow for assimilation or ban for assimilation, you can add that check in config.py (there is a special section for checks on the config file).

googlehydrology/evaluation/tester.py
```diff
@@ -657,14 +667,21 @@ def _evaluate(
...
     ):
         predict_last_n = self.cfg.predict_last_n
         if isinstance(predict_last_n, int):
             predict_last_n = {
                 frequencies[0]: predict_last_n
             }  # if predict_last_n is int, there's only one frequency
 
-        with torch.inference_mode():
+        if data_assimilation:
+            assimilation = Assimilation(self.cfg.assimilation_config)
```
grey-nearing, line 679: Is there a check whether the assimilation_config argument is present in the config file if the cli flag for assimilation is used?

sylvesterkaczmarek (commented):
Precipitation DA silently falls back to the first hindcast feature when no precipitation key matches. That can turn `assimilation_targets=['precip']` into optimisation of temperature, radiation, or another forcing while still producing a plausible rollout. Could this fail explicitly, or require a configured forcing key, instead of selecting `list(hind_dict.keys())[0]`? For DA this seems safer than silently changing the physical variable being assimilated.

kruparell:
> Precipitation DA silently falls back to the first hindcast feature when no precipitation key matches. That can turn `assimilation_targets=['precip']` into optimisation of temperature, radiation, or another forcing while still producing a plausible rollout. Could this fail explicitly, or require a configured forcing key, instead of selecting `list(hind_dict.keys())[0]`? For DA this seems safer than silently changing the physical variable being assimilated.

Thanks for calling this out! You're completely right. Falling back to list(hind_dict.keys())[0] was a silent bug that could optimize radiation or temperature when precipitation keys didn't match.

I have pushed an update that:

Optimizes all matching precipitation features (hres_total_precipitation, graphcast_total_precipitation, imerg_precipitation, cpc_precipitation) together.
Raises an explicit ValueError listing available keys if no precipitation key is found, eliminating the silent fallback.
Supports optional explicit configuration via precip_forcing_keys or precip_forcing_key in AssimilationConfig.

googlehydrology/evaluation/assimilation.py
```diff
@@ -0,0 +1,666 @@
...
```
grey-nearing: Same as above.

googlehydrology/evaluation/assimilation.py
```diff
@@ -0,0 +1,666 @@
...
```
grey-nearing: Same as above.

googlehydrology/evaluation/assimilation.py
```diff
@@ -0,0 +1,666 @@
...
+        elif isinstance(v, dict):
+            sub_len = _infer_seq_len(v)
+            if sub_len is not None:
+                return sub_len
+    return None
+
+
+def _get_var_lr(lr_cfg: Any, var_name: str) -> float:
+    """Retrieves target-specific learning rate from lr_cfg."""
+    if isinstance(lr_cfg, dict):
```
grey-nearing: Outside of DA, this dict is typically meant to be {epoch_int: rl}.

googlehydrology/training/__init__.py
```diff
@@ -23,55 +23,69 @@
...
+        Optimizer object that can be used for model training or data assimilation.
     """
+    params = model.parameters() if hasattr(model, 'parameters') else model
+
+    # Resolve learning rate
+    lr = getattr(cfg, 'initial_learning_rate', None)
+    if lr is None:
+        if hasattr(cfg, 'learning_rate'):
+            lr_val = cfg.learning_rate
+            lr = lr_val[0] if isinstance(lr_val, dict) else lr_val
```
grey-nearing: Here, we assume that the learning rate dict is `ecpoch`: `rate`. Other places, it looks like it can be `variable`: `rate`.

googlehydrology/evaluation/assimilation.py
```diff
@@ -0,0 +1,666 @@
...
+        """Parses self.targets into boolean flags for (c_hc, h_hc, c_fc, h_fc)."""
+        opt_c_hc = any(k in self.targets for k in ['c_n_hindcast', 'c_0_hindcast', 'c_hc', 'c_n', 'c_0'])
+        opt_h_hc = any(k in self.targets for k in ['h_n_hindcast', 'h_0_hindcast', 'h_hc', 'h_n', 'h_0'])
+        opt_c_fc = any(k in self.targets for k in ['c_n_forecast', 'c_0_forecast', 'c_fc', 'c_n', 'c_0'])
+        opt_h_fc = any(k in self.targets for k in ['h_n_forecast', 'h_0_forecast', 'h_fc', 'h_n', 'h_0'])
+        return opt_c_hc, opt_h_hc, opt_c_fc, opt_h_fc
+
+    def _detect_da_type(self) -> str:
+        """Determines the DA mode from self.targets: 'embedding', 'precip', or 'state'."""
+        targets = [str(t).lower() for t in self.targets] if isinstance(self.targets, (list, tuple)) else [str(self.targets).lower()]
```
grey-nearing: If self.targets = ['hindcast_embedding'], target_str is 'hindcast_embedding'. Because 'embedding' is in ['stat', 'both', 'all', 'embedding'], 'embedding' in target_str evaluates to True!

googlehydrology/evaluation/assimilation.py
```diff
@@ -0,0 +1,666 @@
...
+        target_str = ' '.join(targets)
+        if any(k in target_str for k in ['embedded', 'embedding', 'static_embedding', 'hindcast_embedding', 'forecast_embedding']):
+            return 'embedding'
+        elif any(k in target_str for k in ['precip', 'precipitation', 'forcing', 'rain', 'tp']):
+            return 'precip'
+        return 'state'
+
+    def _parse_embedding_masks(self) -> Tuple[bool, bool, bool]:
+        """Returns (mask_e_stat, mask_e_dyn, mask_e_fc) based on configured targets."""
+        targets = [str(t).lower() for t in self.targets] if isinstance(self.targets, (list, tuple)) else [str(self.targets).lower()]
```
grey-nearing: Same as above.

notebooks/multimet_helpers.py
grey-nearing, line 1: Did you mean to commit this notebook? It has paths to your local machine.

notebooks/Data_Assimilation.ipynb
grey-nearing, line 1: This notebook also has paths to your personal machine. 

I have not reviewed either of the notebooks yet.

googlehydrology/evaluation/assimilation.py
```diff
@@ -0,0 +1,666 @@
...
+                            p_sub = pred_dict['y_hat'][:, :win_len, :]
+                            if p_sub.ndim == 2: p_sub = p_sub.unsqueeze(-1)
+                            t_sub = chunk_data['y'][:, :win_len, :]
+
+                            mask = ~torch.isnan(t_sub) & ~torch.isnan(p_sub)
+                            if mask.any():
+                                loss = torch.mean((p_sub[mask] - t_sub[mask]) ** 2)
+                                reg_loss = 0.0
+                                if mask_e_stat and e_stat_opt.requires_grad:
+                                    reg_loss = reg_loss + bg_stat_w * torch.sum((e_stat_opt - e_stat_base) ** 2)
```
grey-nearing: Why do you sum here instead of mean?

googlehydrology/evaluation/assimilation.py
```diff
@@ -0,0 +1,666 @@
...
+
+                        if not (torch.isnan(val_at_t) or torch.isnan(y_same) or torch.isnan(y_prev)):
+                            if torch.abs(val_at_t - y_same).item() < 1e-5 and torch.abs(val_at_t - y_prev).item() > 1e-4:
+                                diagnostics['has_timing_mismatch'] = True
+                                diagnostics['warnings'].append(f"TIMING MISMATCH DETECTED in '{feat_name}' at t={check_t}.")
+        return diagnostics
+
+    def _parse_target_flags(self) -> Tuple[bool, bool, bool, bool]:
+        """Parses self.targets into boolean flags for (c_hc, h_hc, c_fc, h_fc)."""
+        opt_c_hc = any(k in self.targets for k in ['c_n_hindcast', 'c_0_hindcast', 'c_hc', 'c_n', 'c_0'])
```
grey-nearing: Can we use more descriptive variable names? Or at least have a comment that describes the variabel naming convention.

README.md
```diff
@@ -110,6 +110,12 @@ Calculate performance metrics (NSE, KGE) on the test set:
    run evaluate --run-dir /path/to/your/model_run/
    ```
 
+To run evaluation with test-time 4D-Var Data Assimilation:
+
+   ```
+   run evaluate --run-dir /path/to/your/model_run/ --data-assimilation
```
grey-nearing, line 116: Can we use inference mode with DA as well? The difference between evaluate vs. infer is that infer saves the timeseries output whereas evaluate only saves the performance metrics.

googlehydrology/training/__init__.py
```diff
@@ -23,55 +23,69 @@
 LOGGER = logging.getLogger(__name__)
 
 
+from typing import Iterable, Union
+
+
 def get_optimizer(
-    model: torch.nn.Module, cfg: Config, *, is_gpu: bool = False
+    model: Union[torch.nn.Module, Iterable[torch.Tensor]], cfg: Config, *, is_gpu: bool = False
```
grey-nearing: Flagging this for myself. Why do we need to allow a different model type?

grey-nearing (commented):
Temp & partial review while you reconfigure the PRs.

googlehydrology/evaluation/assimilation.py
```diff
@@ -0,0 +1,499 @@
...
+            # =========================================================================
+            # PHASE 2: Sequential Window Optimization (a_start -> a_end)
+            # =========================================================================
+            curr_idx = a_start
+            for w_idx in range(self.history):
+                if curr_idx >= a_end: break
+                w_end = min(curr_idx + self.window, a_end)
+                win_len = w_end - curr_idx
+
+                chunk_data = _slice_hydrology_batch(data, curr_idx, w_end)
```
grey-nearing: As discussed in person, the process of choosing the assimilation update targets should be model agnostic. I recommend having each model include a required attribute that lists the model components available for assimilation updating, and then require the user to list the components they want assimilation to target int he config file (with appropriate existence checks). Then direct the optimizer to the selected components generically, in a model-agnostic way.

grey-nearing (commented):
more temp comments.

googlehydrology/evaluation/tester.py
kruparell: I've added this in now

README.md
kruparell: Yes we can, I've added a description of this to the README

sylvesterkaczmarek (commented):
`assimilation_lead_time > seq_length` is not rejected here. In that case `assimilation_end_step` becomes negative, the history loop exits immediately, and the post-DA branch runs the full sequence, so an invalid DA configuration silently behaves like a baseline forecast. Could we validate `0 <= assimilation_lead_time <= seq_length` when building the config/Assimilation and add a regression for the overlong lead-time case?

PR #287, opened 2026-09-03
kruparell-data-assimilation  [4/4]: Add latent embedding space data assimilation

kruparell:


sylvesterkaczmarek (commented):
`_parse_embedding_masks()` makes `mask_e_stat` true for every embedding target because it matches the generic substring `embedding`. For example, `forecast_embedding` enables both static and forecast optimisation, while `hindcast_embedding` enables static and hindcast. Please match the specific target names rather than using the generic substring for the static mask.

PR #288, opened 2026-09-08
Fix HandoffForecastLSTM feature ordering

### LINKED ISSUE #281: [Bug] HandoffForecastLSTM: Silent covariate shift caused by dictionary iteration order in feature concatenation
### Description                                                                                                                                                                                                                                        
In `googlehydrology.modelzoo.handoff_forecast_lstm.HandoffForecastLSTM`, feature concatenation relies on dictionary iteration order rather than the model's fixed feature list order. This creates a silent covariate shift risk where feature columns 
  can be permuted without raising PyTorch dimension errors.                                                                                                                                                                                                
                                                                                                                                                                                                                                                              
### Locations                                                                                                                                                                                                                                          
`googlehydrology/modelzoo/handoff_forecast_lstm.py`:                                                                                                                                                                                                      
- `forward()` (lines 260–269)                                                                                                                                                                                                                             
- `save_state()` (lines 467–474)        
- `predict_from_state()` (lines 490–497)
         
```python
hindcast_features = torch.cat(
    [
        t for f, t in data['x_d_hindcast'].items()
        if f in self.hindcast_inputs
    ], dim=-1)
forecast_features = torch.cat(
    [
        t for f, t in data['x_d_forecast'].items()
        if f in self.forecast_inputs
    ], dim=-1)
```
### Failure Mechanism       

  1. data['x_d_hindcast'].items() yields key-tensor pairs in the arbitrary insertion order of the input dictionary.
  2. The model's weights in self.hindcast_embedding_net expect an immutable feature column order defined at initialization (e.g., column 0 = precipitation, column 1 = temperature).
  3. If a different dataloader, dataset format (e.g. NetCDF vs Zarr), or pipeline creates data['x_d_hindcast'] with keys in a different order, columns swap positions (e.g., precipitation inputs get fed into temperature weights).
  4. Because the tensor dimensions ([batch, seq_len, n_features]) remain identical, PyTorch raises no error, leading to silent model degradation.

  ### Reference Implementation

  MeanEmbeddingForecastLSTM avoids this bug (googlehydrology/modelzoo/mean_embedding_forecast_lstm.py:553-556) by explicitly indexing the dictionary using the model's key sequence:

    def _concat_tensors_from_dict(data: dict[str, torch.Tensor], *, keys: Iterable[str]) -> torch.Tensor:
        return torch.cat([data[e] for e in keys], dim=-1)

  ### Proposed Fix

  Iterate over self.hindcast_inputs and self.forecast_inputs instead of .items():

```python
    hindcast_features = torch.cat(
        [
            data['x_d_hindcast'][f]      
            for f in self.hindcast_inputs
            if f in data['x_d_hindcast']
        ],     
        dim=-1,
    )     

    forecast_features = torch.cat(
        [
            data['x_d_forecast'][f]      
            for f in self.forecast_inputs
            if f in data['x_d_forecast']
        ],     
        dim=-1,
    )

### PR DESCRIPTION
**Pull request by sylvesterkaczmarek**
Fixes #281.

Concatenates hindcast and forecast dynamic features in the model-defined feature order instead of dictionary insertion order, including the hot-start state-saving path. Adds a regression test with deliberately reversed dictionary insertion order.

Validation:
- targeted feature-order regression check
- Python syntax compilation
- git diff --check

### DIFF
--- a/googlehydrology/modelzoo/handoff_forecast_lstm.py
+++ b/googlehydrology/modelzoo/handoff_forecast_lstm.py
@@ -28,6 +28,13 @@ from googlehydrology.utils.lstm_utils import lstm_init
 FC_XAVIER = WeightInitOpt.FC_XAVIER
 
 
+def _concat_dynamic_features(
+    data: dict[str, torch.Tensor], feature_names: list[str]
+) -> torch.Tensor:
+    """Concatenate dynamic features in the model-defined feature order."""
+    return torch.cat([data[name] for name in feature_names], dim=-1)
+
+
 class HandoffForecastLSTM(BaseModel):
     """
     An encoder/decoder LSTM model class used for forecasting.
@@ -247,16 +254,12 @@ class HandoffForecastLSTM(BaseModel):
         """
 
         # Run the embedding layers.
-        hindcast_features = torch.cat(
-            [
-                t for f, t in data['x_d_hindcast'].items()
-                if f in self.hindcast_inputs
-            ], dim=-1)
-        forecast_features = torch.cat(
-            [
-                t for f, t in data['x_d_forecast'].items()
-                if f in self.forecast_inputs
-            ], dim=-1)
+        hindcast_features = _concat_dynamic_features(
+            data['x_d_hindcast'], self.hindcast_inputs
+        )
+        forecast_features = _concat_dynamic_features(
+            data['x_d_forecast'], self.forecast_inputs
+        )
 
         statics_embeddings = self.statics_embedding_net(data['x_s'])
         hindcast_embeddings = self.hindcast_embedding_net(hindcast_features)
@@ -454,13 +457,8 @@ class HandoffForecastLSTM(BaseModel):
             The file path where the state should be saved (.npz format).
         """
         # Run the embedding layers.
-        hindcast_features = torch.cat(
-            [
-                t
-                for f, t in data['x_d_hindcast'].items()
-                if f in self.hindcast_inputs
-            ],
-            dim=-1,
+        hindcast_features = _concat_dynamic_features(
+            data['x_d_hindcast'], self.hindcast_inputs
         )
 
         statics_embeddings = self.statics_embedding_net(data['x_s'])
@@ -477,13 +475,8 @@ class HandoffForecastLSTM(BaseModel):
         )
 
         # We run the exact same logic up to the final temporal state (Day D)
-        forecast_features = torch.cat(
-            [
-                t
-                for f, t in data['x_d_forecast'].items()
-                if f in self.forecast_inputs
-            ],
-            dim=-1,
+        forecast_features = _concat_dynamic_features(
+            data['x_d_forecast'], self.forecast_inputs
         )
 
         forecast_embeddings = self.forecast_embedding_net(forecast_features)
--- a/test/test_handoff_forecast_lstm.py
+++ b/test/test_handoff_forecast_lstm.py
@@ -1,0 +1,35 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Tests for HandoffForecastLSTM feature ordering."""
+
+import torch
+
+from googlehydrology.modelzoo.handoff_forecast_lstm import (
+    _concat_dynamic_features,
+)
+
+
+def test_concat_dynamic_features_uses_model_feature_order() -> None:
+    data = {
+        'temperature': torch.tensor([[[2.0], [4.0]]]),
+        'precipitation': torch.tensor([[[1.0], [3.0]]]),
+    }
+
+    result = _concat_dynamic_features(
+        data, ['precipitation', 'temperature']
+    )
+
+    expected = torch.tensor([[[1.0, 2.0], [3.0, 4.0]]])
+    assert torch.equal(result, expected)

### CODE REVIEW

File: test/test_handoff_forecast_lstm.py [hunk 1]
```diff
@@ -0,0 +1,35 @@
...
+        'temperature': torch.tensor([[[2.0], [4.0]]]),
+        'precipitation': torch.tensor([[[1.0], [3.0]]]),
+    }
+
+    result = _concat_dynamic_features(
+        data, ['precipitation', 'temperature']
+    )
+
+    expected = torch.tensor([[[1.0, 2.0], [3.0, 4.0]]])
+    assert torch.equal(result, expected)
```
grey-nearing: **This test doesn't actually test the regression.** 🔴

It calls the new private helper directly, which asserts that a two-line function does what its body plainly says. If someone reverted `forward()` to the old `.items()` code, this test would still pass — so it provides no protection against the bug it's meant to lock down. It also couples the suite to a private symbol (`_concat_dynamic_features`), pinning an implementation detail.

Could you replace it with a test that exercises the actual model? Build a small `HandoffForecastLSTM`, run it twice with the same tensors in two different dict insertion orders, and assert the outputs match. `get_base_cfg` in `test/test_hot_start.py` is reusable for the config. Something like:

```python
@patch('googlehydrology.datautils.scaler.Scaler.load')
@patch('googlehydrology.datautils.scaler.Scaler.check_zero_scale')
def test_forward_invariant_to_feature_dict_order(mock_check, mock_load, tmp_path):
    """Feature columns must follow cfg order, not dict insertion order (#281)."""
    cfg = Config(get_base_cfg(tmp_path), dev_mode=True)
    model = HandoffForecastLSTM(cfg)
    model.eval()

    data = _make_data(cfg)  # dict built in cfg order
    shuffled = {
        'x_d_hindcast': {k: data['x_d_hindcast'][k]
                         for k in reversed(list(data['x_d_hindcast']))},
        'x_d_forecast': {k: data['x_d_forecast'][k]
                         for k in reversed(list(data['x_d_forecast']))},
        'x_s': data['x_s'],
    }

    with torch.no_grad():
        assert torch.allclose(
            model(data)['y_hat'], model(shuffled)['y_hat'], atol=1e-6
        )
```

Please add the equivalent for `save_state()` too (run it twice and compare the two `.npz` archives) — that path is independently fixed here and deserves its own guard.

I ran both versions locally: they fail on `main` and pass on this branch, which is what we want out of a regression test. A third cheap case worth adding is asserting that a feature missing from the batch now raises, so the stricter behavior is intentional and pinned rather than incidental.

googlehydrology/modelzoo/handoff_forecast_lstm.py
```diff
@@ -28,6 +28,13 @@
 FC_XAVIER = WeightInitOpt.FC_XAVIER
 
 
+def _concat_dynamic_features(
+    data: dict[str, torch.Tensor], feature_names: list[str]
+) -> torch.Tensor:
+    """Concatenate dynamic features in the model-defined feature order."""
+    return torch.cat([data[name] for name in feature_names], dim=-1)
```
grey-nearing, line 40: Two notes on the helper itself.

**1. Please don't duplicate it.** `mean_embedding_forecast_lstm.py:553` already has this exact function under a different name, and #281 cites it as the reference implementation:

```python
def _concat_tensors_from_dict(data: dict[str, torch.Tensor], *, keys: Iterable[str]) -> torch.Tensor:
    return torch.cat([data[e] for e in keys], dim=-1)
```

Two near-identical private helpers in sibling modules means the next model added to `modelzoo/` gets a coin flip as to which one it copies. Ideally hoist one shared helper (e.g. into `googlehydrology/modelzoo/` as a shared util) and have both import it. If you'd rather keep this PR surgical I'm fine with a follow-up issue — but at minimum please match the twin's signature (keyword-only `keys`).

**2. Consider an explicit error for missing features.** Right now a feature named in the config but absent from the batch escapes as a bare `KeyError` from inside `torch.cat`, which isn't a great diagnostic. This is arguably the highest-value line in the PR, since it turns the remaining silent-failure mode into a loud one:

```suggestion
def _concat_dynamic_features(
    data: dict[str, torch.Tensor], feature_names: list[str]
) -> torch.Tensor:
    """Concatenate dynamic features in the model-defined feature order."""
    missing = [name for name in feature_names if name not in data]
    if missing:
        raise KeyError(f'Missing dynamic features in batch: {missing}')
    return torch.cat([data[name] for name in feature_names], dim=-1)
```

To be explicit for future readers: raising here is deliberate and is the right call. It differs from the fix proposed in #281, which suggested an `if f in data[...]` guard — a silent skip would just relocate the failure to a less diagnosable shape mismatch downstream.

Worth noting one benign side effect of the new indexing: if a feature name appears twice in `hindcast_inputs` (possible with grouped/dict-style configs), it's now concatenated twice, which actually *matches* the `len(self.hindcast_inputs)` the embedding layer is sized with. The old code deduplicated and crashed. Small improvement.

googlehydrology/modelzoo/handoff_forecast_lstm.py
```diff
@@ -247,16 +254,12 @@ def forward(self, data: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
...
-            [
-                t for f, t in data['x_d_forecast'].items()
-                if f in self.forecast_inputs
-            ], dim=-1)
+        hindcast_features = _concat_dynamic_features(
+            data['x_d_hindcast'], self.hindcast_inputs
+        )
+        forecast_features = _concat_dynamic_features(
+            data['x_d_forecast'], self.forecast_inputs
+        )
```
grey-nearing, line 267: Confirming this is correct and is the behavioral heart of the fix.

`self.hindcast_inputs` / `self.forecast_inputs` come from `flatten_feature_list(cfg.*)`, which is the same function `Multimet` uses to build `_hindcast_features` — so the names are guaranteed to line up, and the embedding layers are already sized with `len(self.hindcast_inputs)`. Indexing by the model's list is the correct source of truth.

I checked the blast radius for the stricter lookup: `Multimet` is the only in-repo producer of `x_d_hindcast` / `x_d_forecast`, and `_extract_hindcasts` / `_extract_forecasts` always return exactly the configured feature set, so nothing in `googlehydrology/` or `tools/` can hit the new `KeyError` path. Extra keys the dataset adds (`hindcast_counter`, `forecast_counter` under `timestep_counter: true`) are still excluded, same as before.

grey-nearing (changes requested):
Thanks for this — the diagnosis in #281 is accurate and the fix is the right one.

I want to record that I verified this is a **live bug, not a theoretical one**. `Multimet` partitions hindcast features by whether the underlying dataset variable carries a `lead_time` dimension and then merges the two groups (`multimet.py:237-246`, `455-473`), so the batch order is a function of *the data*, not the config. With our own `example-configs/handoff-forecast-lstm-config.yml` that means:

| | Order |
|---|---|
| Config declares | `hres_snsr, hres_sntr, hres_sp, hres_t2m, hres_tp, imerg, cpc` |
| Old code fed the network | `imerg, cpc, hres_snsr, hres_sntr, hres_sp, hres_t2m, hres_tp` |

I reproduced that directly against a real `Multimet` instance. Because the partition is data-dependent, the columns can also silently permute between two stores of the same config (e.g. HRES with vs. without a materialized `lead_time` dim) — which is exactly the silent covariate shift #281 describes.

I also confirmed the fix is complete: all four dict-order concatenations in this file (two in `forward()`, two in `save_state()`) are covered, with none left behind. The issue text mentions `predict_from_state()`, but that method no longer exists on `main`.

Verification I ran locally (torch 2.5.1):
- Full suite `pytest -m "not gpu"`: **219 passed**, 0 failed.
- A purpose-built order-invariance test for `forward()` and `save_state()`: **fails on `main`, passes on this branch.**

Two requests before merge, then I'm happy to approve — see the inline comments.

---

⚠️ **Note for merge (not the author's responsibility):** this changes the feature column order, so any `HandoffForecastLSTM` checkpoint trained before this merge will silently change its predictions and needs retraining. The two pretrained models shipped in this repo are `mean_embedding_forecast_lstm`, which already used the safe path, so they are unaffected. I'll add a release note.

Also: `pytest CI` has not run on this PR (fork PRs need "Approve and run workflows"), which is why it shows as `unstable`. I'll kick that off.

test/test_handoff_forecast_lstm.py
sylvesterkaczmarek: Addressed in `774ca5f99e6c71465ebec07f296520ece05dd60f`. Replaced the helper-only test with end-to-end `HandoffForecastLSTM` regressions: `forward()` is invariant to dynamic-feature dict insertion order, `save_state()` produces identical archives across insertion orders, and a configured feature missing from the batch raises explicitly. The focused test file passes 3/3 locally.

googlehydrology/modelzoo/handoff_forecast_lstm.py
sylvesterkaczmarek: Also addressed in `774ca5f99e6c71465ebec07f296520ece05dd60f`: the helper now matches the sibling keyword-only `keys` signature and reports the missing configured feature names explicitly rather than surfacing a bare indexing error. I kept the helper local to keep this PR focused.

grey-nearing (approved):
Approving — both requests are addressed, thanks for the quick turnaround.

sylvesterkaczmarek:
Thanks for the review and merge, and for validating the feature-order bug against a real Multimet setup. I appreciate the thorough regression check.

PR #289, opened 2026-09-08
Fix LSTM forget bias initialization

### LINKED ISSUE #278: [Robustness] `lstm_init` only sets forget bias for layer 0 and raises AttributeError on `bias=False`
### Description

In `googlehydrology/utils/lstm_utils.py:38-39`, `lstm_init` hardcodes access to `lstm.bias_hh_l0`:

```python
if forget_bias is not None:
    lstm.bias_hh_l0.data[_forget_gate_slice(lstm)] = forget_bias
```

While existing models in the repository currently instantiate single-layer LSTMs with bias=True, the utility fails when used with other standard PyTorch LSTM configurations:

1. Multi-layer LSTMs (num_layers > 1): Weight initialization options (lines 41-45) correctly iterate over all layers, but forget_bias is only applied to layer 0 (bias_hh_l0). Higher layers (bias_hh_l1, etc.) remain uninitialized with custom forget bias.
2. Unbiased LSTMs (bias=False): PyTorch does not instantiate bias attributes when bias=False. Calling lstm_init on an unbiased LSTM raises AttributeError: 'LSTM' object has no attribute 'bias_hh_l0'.

### Affected Files

• googlehydrology/utils/lstm_utils.py (lines 38-39)

### Proposed Fix

Iterate over lstm.num_layers and check lstm.bias:
```
if forget_bias is not None and lstm.bias:
    for layer in range(lstm.num_layers):
        bias_param = getattr(lstm, f'bias_hh_l{layer}', None)
        if bias_param is not None:
            bias_param.data[_forget_gate_slice(lstm)] = forget_bias
```

sylvesterkaczmarek:
Fixes #278.

Initialises the configured forget bias for every recurrent hidden-bias parameter instead of only `bias_hh_l0`. This covers multi-layer and bidirectional LSTMs, while `bias=False` now works naturally because no bias parameters are present.

Adds regression coverage for all layers/directions and for unbiased LSTMs.

Validation:
- `python -m pytest -q test/test_lstm_utils.py` (3 passed)
- `ruff check googlehydrology/utils/lstm_utils.py test/test_lstm_utils.py`
- Python syntax compilation
- `git diff --check`

### DIFF
--- a/googlehydrology/utils/lstm_utils.py
+++ b/googlehydrology/utils/lstm_utils.py
@@ -35,11 +35,10 @@ def lstm_init(
     """Initialize LSTM weights."""
     with torch.no_grad():
         for lstm in lstms:
-            if forget_bias is not None:
-                lstm.bias_hh_l0.data[_forget_gate_slice(lstm)] = forget_bias
-
             for name, param in lstm.named_parameters():
-                if 'weight_ih' in name and LSTM_IH_XAVIER in weight_opts:
+                if forget_bias is not None and name.startswith('bias_hh_'):
+                    param.data[_forget_gate_slice(lstm)] = forget_bias
+                elif 'weight_ih' in name and LSTM_IH_XAVIER in weight_opts:
                     nn.init.xavier_uniform_(param)
                 elif 'weight_hh' in name and LSTM_HH_ORTHOGONAL in weight_opts:
                     nn.init.orthogonal_(param)
--- a/test/test_lstm_utils.py
+++ b/test/test_lstm_utils.py
@@ -31,7 +31,9 @@ def test_forget_gate_slice():
 
 @pytest.mark.unit
 def test_lstm_init():
-    lstm = nn.LSTM(input_size=10, hidden_size=20)
+    lstm = nn.LSTM(
+        input_size=10, hidden_size=20, num_layers=3, bidirectional=True
+    )
     lstm_init(
         lstms=[lstm],
         forget_bias=1.5,
@@ -40,6 +42,14 @@ def test_lstm_init():
             WeightInitOpt.LSTM_HH_ORTHOGONAL,
         ],
     )
-    # Check forget bias initialized
+    # Check forget bias initialized for every layer and direction.
     sl = _forget_gate_slice(lstm)
-    assert torch.allclose(lstm.bias_hh_l0.data[sl], torch.tensor(1.5))
+    for name, param in lstm.named_parameters():
+        if name.startswith('bias_hh_'):
+            assert torch.allclose(param.data[sl], torch.tensor(1.5))
+
+
+@pytest.mark.unit
+def test_lstm_init_without_bias():
+    lstm = nn.LSTM(input_size=10, hidden_size=20, bias=False)
+    lstm_init(lstms=[lstm], forget_bias=1.5)


googlehydrology/utils/lstm_utils.py
```diff
@@ -35,11 +35,10 @@ def lstm_init(
     """Initialize LSTM weights."""
     with torch.no_grad():
         for lstm in lstms:
-            if forget_bias is not None:
-                lstm.bias_hh_l0.data[_forget_gate_slice(lstm)] = forget_bias
-
             for name, param in lstm.named_parameters():
-                if 'weight_ih' in name and LSTM_IH_XAVIER in weight_opts:
+                if forget_bias is not None and name.startswith('bias_hh_'):
+                    param.data[_forget_gate_slice(lstm)] = forget_bias
```
grey-nearing, line 40: Correct, and I confirmed the `elif` restructuring is safe: `bias_hh_*` never matches `'weight_ih'` / `'weight_hh'`, there's no bias branch further down the chain, and bias vs. weight initialization touch disjoint parameters — so moving the forget bias from before the loop to inside it is behaviourally inert. The bit-identical results in my summary table confirm it empirically.

Two optional nits.

**1. Match the file's existing name-matching idiom.** `startswith('bias_hh_')` sits directly above two branches that use `'weight_ih' in name`. Beyond consistency, `in` is more forgiving: if these parameters ever arrive prefixed (someone passing a wrapper module rather than a bare `nn.LSTM`), `startswith` silently matches nothing and the forget bias is quietly skipped — whereas the old code would at least have failed loudly.

```suggestion
            for name, param in lstm.named_parameters():
                if forget_bias is not None and 'bias_hh' in name:
                    param.data[_forget_gate_slice(lstm)] = forget_bias
```

**2. Consider keeping the two concerns decoupled (non-blocking).** The `elif` chain now makes the weight-init branches conditional on the bias branch not matching. That's invisible coupling for whoever later adds a bias-init option, and the precedence would be decided silently by branch order. An alternative:

```python
if forget_bias is not None:
    for name, param in lstm.named_parameters():
        if 'bias_hh' in name:
            param.data[_forget_gate_slice(lstm)] = forget_bias

for name, param in lstm.named_parameters():
    if 'weight_ih' in name and LSTM_IH_XAVIER in weight_opts:
        ...
```

Two passes over a handful of parameters at init time costs nothing. Entirely your call — the current form is correct and I'm happy to merge as-is.

test/test_lstm_utils.py
```diff
@@ -40,6 +42,14 @@ def test_lstm_init():
...
-    assert torch.allclose(lstm.bias_hh_l0.data[sl], torch.tensor(1.5))
+    for name, param in lstm.named_parameters():
+        if name.startswith('bias_hh_'):
+            assert torch.allclose(param.data[sl], torch.tensor(1.5))
+
+
+@pytest.mark.unit
+def test_lstm_init_without_bias():
+    lstm = nn.LSTM(input_size=10, hidden_size=20, bias=False)
+    lstm_init(lstms=[lstm], forget_bias=1.5)
```
grey-nearing, line 55: Good test — this is exactly the right shape for pinning the `AttributeError` regression, and I verified it fails on `main` and passes here.

One small thing: it passes purely by *not raising*, with no assertion, so it reads like an unfinished test to anyone skimming it later. A single line makes the intent explicit:

```suggestion
@pytest.mark.unit
def test_lstm_init_without_bias() -> None:
    lstm = nn.LSTM(input_size=10, hidden_size=20, bias=False)
    lstm_init(lstms=[lstm], forget_bias=1.5)  # must not raise
    assert not any(n.startswith('bias') for n, _ in lstm.named_parameters())
```

That also adds the `-> None` annotation you used over in #288 (`ruff` flags `ANN201`/`D103` here). Purely cosmetic — it matches the surrounding file's existing style and ruff isn't wired into CI, so ignore if you'd rather not churn it.

Separately, nice job upgrading `test_lstm_init` to `num_layers=3, bidirectional=True` and looping over every `bias_hh_*` — that covers all six parameters without adding a new test.

grey-nearing:
AI review approved by me.

grey-nearing (approved):
Thanks — this one's clean, and I'm happy to approve it.

Nice call driving the fix off `named_parameters()` rather than the `range(num_layers)` + `getattr` approach suggested in #278: that version would have silently missed every `bias_hh_l*_reverse` parameter on bidirectional LSTMs, so your fix is strictly more complete than the one in the issue.

**Blast radius — verified, not assumed.** I diffed old-vs-new initialization from identical seeds across layer counts, directions, `bias=False`, and `proj_size`:

| Configuration | forget bias applied (old → new) | Result |
|---|---|---|
| **1 layer, bias=True** ← every LSTM in this repo | 1/1 → 1/1 | **bit-identical** |
| 3 layers | 1/3 → 3/3 | differs: `bias_hh_l1`, `bias_hh_l2` |
| bidirectional | 1/2 → 2/2 | differs: `bias_hh_l0_reverse` |
| 3 layers + bidirectional | 1/6 → 6/6 | differs: 5 params |
| `bias=False` | `AttributeError` → ok | bug fixed |
| `proj_size=3` | 1/1 → 1/1 | bit-identical |

All four LSTMs in the codebase (`hindcast_lstm` / `forecast_lstm` in both `handoff_forecast_lstm.py` and `mean_embedding_forecast_lstm.py`) are single-layer, unidirectional, `bias=True`, so they land on the bit-identical row. **No existing run or checkpoint changes**, and the differences appear only in the configurations that were broken. No release note needed.

**The tests are real regression tests**, which I appreciate — I ran them against unpatched `main` and both fail there (`bias_hh_l0_reverse` still random; `AttributeError` for `bias=False`), then pass on this branch. Full suite on the patched tree: 219 passed.

Approving. Two optional nits inline that I'm fine seeing deferred to a follow-up.

---

One out-of-scope observation, flagged for tracking rather than for this PR: PyTorch computes the forget gate as `σ(W_if x + b_if + W_hf h + b_hf)`, so **both** bias terms contribute. We only set `b_hf`, leaving `b_if` at its default `U(-1/√H, 1/√H)` — meaning the effective initial forget bias is `forget_bias ± 1/√hidden_size` rather than exactly the configured value (≈ ±0.06 at `hidden_size=256`). That matches upstream NeuralHydrology, so it's almost certainly inherited intent rather than a defect, but since `initial_forget_bias` is a knob people deliberately tune for long-memory catchments, I'll open a separate issue to decide whether it should be exactly realized.

Note for whoever merges: `pytest CI` hasn't run here either (fork PRs need "Approve and run workflows") — I'll kick it off.

sylvesterkaczmarek:
Thanks for the careful verification and approval, especially checking the blast radius across layer count, direction, bias, and projection settings. I also appreciate you kicking off the fork CI.

googlehydrology/utils/lstm_utils.py
grey-nearing: No change needed — resolving. Both points were optional and the current form is correct; not worth another CI cycle for cosmetics.

test/test_lstm_utils.py
grey-nearing: No change needed — resolving. Passing by not raising is a fine way to pin the AttributeError regression.

PR #290, opened 2026-09-10
Fix CMAL loss masking for missing timesteps

### LINKED ISSUE #275: [BUG] MaskedCMALLoss drops entire sequence on any single NaN timestep instead of elementwise masking
**Location**
`googlehydrology/training/loss.py:403-408`

**Root Cause Mechanism**
Evaluating `mask = ~torch.isnan(ground_truth['y']).any(1).any(1)` collapses sequences into a 1D batch-level mask. If a single timestep in a sequence (e.g., 1 day out of 365) contains a missing value (`NaN`), the entire basin trajectory is dropped. When all sequences in a mini-batch contain at least one missing observation, `mask` evaluates to all-`False`, triggering empty tensor reductions and yielding `NaN` loss gradients.

**Impact**
Substantially degrades CMAL training efficiency on operational gauge datasets with intermittent missing observations by dropping valid discharge timesteps, while causing training crashes when batches contain no completely gap-free sequences.

**Suggested Fix**
Adopt pointwise boolean masking analogous to `MaskedMSELoss`, evaluating per-timestep log-likelihoods and normalizing strictly by the count of observed targets:

```python
valid_mask = ~torch.isnan(ground_truth['y'])  # shape: [batch, seq_len, target_dim]
y = torch.where(valid_mask, ground_truth['y'], torch.zeros_like(ground_truth['y']))

# ... compute mixture log_weights and log_like using sanitized y ...

result = torch.logsumexp(log_weights + log_like, dim=-1, keepdim=True)
valid_count = torch.clamp(valid_mask.sum(), min=1.0)

loss = -torch.sum(result * valid_mask.float()) / valid_count
return loss

```

**Verification Checklist**

* [ ] Test loss computation on synthetic batches with 10%, 50%, and 100% missing sequence entries.
* [ ] Check gradient propagation to ensure zero loss gradients flow from masked `NaN` positions.
* [ ] Validate that loss magnitude remains scale-invariant across batches with varying proportions of missing data.

sylvesterkaczmarek:
Fixes #275.

Masks CMAL likelihoods per observed timestep instead of dropping an entire sequence when one target is missing. The loss is averaged over observed timesteps, and all-missing windows return a differentiable zero instead of NaN.

Adds regression coverage for partial missing targets, all-missing windows, and the crash case where every sequence has at least one missing timestep.

Validation:
- Python syntax compilation
- git diff --check
- hosted cross-platform PyTest CI

### DIFF
--- a/googlehydrology/training/loss.py
+++ b/googlehydrology/training/loss.py
@@ -400,8 +400,12 @@ class MaskedCMALLoss(BaseLoss):
         ground_truth: dict[str, torch.Tensor],
         **kwargs,
     ):
-        mask = ~torch.isnan(ground_truth['y']).any(1).any(1)
-        y = ground_truth['y'][mask]
+        y = ground_truth['y'].squeeze(-1)
+        mask = ~torch.isnan(y)
+        if not torch.any(mask):
+            return prediction['mu'].sum() * 0.0
+
+        y = y[mask].unsqueeze(-1)
         m = prediction['mu'][mask]
         b = prediction['b'][mask]
         t = prediction['tau'][mask]
@@ -416,9 +420,8 @@ class MaskedCMALLoss(BaseLoss):
         )
         log_weights = torch.log(p + self.eps)
 
-        result = torch.logsumexp(log_weights + log_like, dim=2)
-        result = -torch.mean(torch.sum(result, dim=1))
-        return result
+        result = torch.logsumexp(log_weights + log_like, dim=-1)
+        return -torch.mean(result)
 
 
 def _get_predict_last_n(cfg: Config) -> dict:
--- a/test/test_cmal_loss_missing_values.py
+++ b/test/test_cmal_loss_missing_values.py
@@ -1,0 +1,69 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from unittest.mock import MagicMock
+
+import pytest
+import torch
+
+from googlehydrology.training.loss import MaskedCMALLoss
+
+
+def _config() -> MagicMock:
+    cfg = MagicMock()
+    cfg.predict_last_n = 2
+    cfg.no_loss_frequencies = []
+    cfg.target_variables = ['streamflow']
+    cfg.target_loss_weights = None
+    cfg.n_distributions = 3
+    return cfg
+
+
+@pytest.mark.unit
+def test_cmal_loss_masks_individual_missing_timesteps():
+    loss_fn = MaskedCMALLoss(_config())
+
+    mu = torch.zeros(2, 2, 3)
+    b = torch.ones(2, 2, 3)
+    tau = torch.full((2, 2, 3), 0.5)
+    pi = torch.full((2, 2, 3), 1.0 / 3)
+    y = torch.tensor([[[0.0], [torch.nan]], [[2.0], [0.0]]])
+
+    total_loss, _ = loss_fn(
+        {'mu': mu, 'b': b, 'tau': tau, 'pi': pi}, {'y': y}
+    )
+
+    expected = torch.log(torch.tensor(4.0)) + 1.0 / 3.0
+    assert torch.allclose(total_loss, expected, atol=1e-6)
+
+
+@pytest.mark.unit
+def test_cmal_loss_all_missing_is_differentiable_zero():
+    loss_fn = MaskedCMALLoss(_config())
+
+    mu = torch.zeros(2, 2, 3, requires_grad=True)
+    prediction = {
+        'mu': mu,
+        'b': torch.ones(2, 2, 3),
+        'tau': torch.full((2, 2, 3), 0.5),
+        'pi': torch.full((2, 2, 3), 1.0 / 3),
+    }
+    y = torch.full((2, 2, 1), torch.nan)
+
+    total_loss, _ = loss_fn(prediction, {'y': y})
+    total_loss.backward()
+
+    assert torch.isfinite(total_loss)
+    assert total_loss.item() == 0.0
+    assert torch.equal(mu.grad, torch.zeros_like(mu))


googlehydrology/training/loss.py
```diff
@@ -416,9 +420,8 @@ def _get_loss(
         )
         log_weights = torch.log(p + self.eps)
 
-        result = torch.logsumexp(log_weights + log_like, dim=2)
-        result = -torch.mean(torch.sum(result, dim=1))
-        return result
+        result = torch.logsumexp(log_weights + log_like, dim=-1)
+        return -torch.mean(result)
```
grey-nearing: 🟡 **This is the line that silently rescales the loss.** `main` did
`-mean(sum(result, dim=1))` (sum over timesteps, mean over batch); this
is `-mean(result)` over observed timesteps, i.e. the old value divided
by `predict_last_n` when nothing is missing. I measured exactly
1.000 / 2.000 / 8.000 / 365.000 for `predict_last_n` = 1 / 2 / 8 / 365.

All eight CMAL configs in the repo use `predict_last_n: 8` with
`clip_gradient_norm: 1`, and on a real LSTM+CMAL head (bs=256, T=8) the
pre-clip gradient norm goes from **1.223 (clipped)** to **0.153 (not
clipped)** — so clipping stops engaging and the optimization path
changes even under Adam.

I'm not asking you to change the math — mean-over-observations is what
#275 asked for and it's consistent with `MaskedMSELoss`. I am asking
for it to be stated out loud: please add a line to the PR description,
and extend the class docstring (around line 378) to say the loss is
averaged over *observed* timesteps and returns 0 for an all-missing
window. If you'd rather preserve the old magnitude instead, the
scale-preserving variant is `return -torch.sum(result) / mask.shape[0]`
— but that gives up the scale invariance across missing fractions that
the issue explicitly asked for, so my preference is your version plus
documentation.

googlehydrology/training/loss.py
```diff
@@ -400,8 +400,12 @@ def _get_loss(
         ground_truth: dict[str, torch.Tensor],
         **kwargs,
     ):
-        mask = ~torch.isnan(ground_truth['y']).any(1).any(1)
-        y = ground_truth['y'][mask]
+        y = ground_truth['y'].squeeze(-1)
+        mask = ~torch.isnan(y)
+        if not torch.any(mask):
+            return prediction['mu'].sum() * 0.0
```
grey-nearing: 🟢 Nit, optional. The early return only touches `mu`, so `b`, `tau` and `pi` come back with `grad is None` for an all-missing window (I verified this). Harmless today — we do not use DDP and Adam skips `None` grads — but it would trip `DistributedDataParallel` without `find_unused_parameters=True`, and `0.0 * inf` would be NaN if the head ever emitted a non-finite `mu`. Touching all four tensors costs nothing:

```suggestion
        if not torch.any(mask):
            # Touch every predicted tensor so that all head parameters
            # receive a (zero) gradient for this batch.
            zero = sum(
                prediction[key].sum() for key in ('mu', 'b', 'tau', 'pi')
            )
            return zero * 0.0
```

I applied this locally and `pytest test/test_cmal_loss_missing_values.py test/test_losses.py` still passes (10 passed), with all four tensors getting zero grads.

test/test_cmal_loss_missing_values.py
```diff
@@ -0,0 +1,69 @@
...
+        'pi': torch.full((2, 2, 3), 1.0 / 3),
+    }
+    y = torch.full((2, 2, 1), torch.nan)
+
+    total_loss, _ = loss_fn(prediction, {'y': y})
+    total_loss.backward()
+
+    assert torch.isfinite(total_loss)
+    assert total_loss.item() == 0.0
+    assert torch.equal(mu.grad, torch.zeros_like(mu))
```
grey-nearing: 🟢 Nit, optional, but this is the case I would most want locked down. Neither test covers the actual crash mode from #275: *every* sequence has a gap, but the batch is not all-NaN. In the first test row 1 is fully observed, so `main` returns a finite (wrong) number rather than NaN — the NaN-abort path is currently untested. This extra case fails on `main` with `isfinite(nan)` and passes here (exact value `log 4`); I ran it both ways:

```suggestion
    assert torch.equal(mu.grad, torch.zeros_like(mu))


@pytest.mark.unit
def test_cmal_loss_is_finite_when_every_sequence_has_a_gap():
    loss_fn = MaskedCMALLoss(_config())

    mu = torch.zeros(2, 2, 3)
    b = torch.ones(2, 2, 3)
    tau = torch.full((2, 2, 3), 0.5)
    pi = torch.full((2, 2, 3), 1.0 / 3)
    # Every sequence has one gap, so the old batch-level mask was
    # all-False and the loss came out NaN, which aborts training
    # (allow_subsequent_nan_losses defaults to 0).
    y = torch.tensor([[[0.0], [torch.nan]], [[torch.nan], [0.0]]])

    total_loss, _ = loss_fn({'mu': mu, 'b': b, 'tau': tau, 'pi': pi}, {'y': y})

    assert torch.isfinite(total_loss)
    assert torch.allclose(total_loss, torch.log(torch.tensor(4.0)), atol=1e-6)
```

test/test_cmal_loss_missing_values.py
```diff
@@ -0,0 +1,69 @@
...
+
+    mu = torch.zeros(2, 2, 3)
+    b = torch.ones(2, 2, 3)
+    tau = torch.full((2, 2, 3), 0.5)
+    pi = torch.full((2, 2, 3), 1.0 / 3)
+    y = torch.tensor([[[0.0], [torch.nan]], [[2.0], [0.0]]])
+
+    total_loss, _ = loss_fn(
+        {'mu': mu, 'b': b, 'tau': tau, 'pi': pi}, {'y': y}
+    )
```
grey-nearing: 🟢 Purely cosmetic and ruff is not in CI, so ignore if you like — `ruff format` collapses this to one line (79 chars, within the project's limit):

```suggestion
    total_loss, _ = loss_fn({'mu': mu, 'b': b, 'tau': tau, 'pi': pi}, {'y': y})
```

googlehydrology/training/loss.py
grey-nearing: **Please add a line to the PR description and to the class docstring saying that this changes the loss scale by a factor of `predict_last_n`.**

No maths change requested. This is the line that does it: `main` did
`-mean(sum(result, dim=1))` (sum over timesteps, mean over batch), and this is
`-mean(result)` over observed timesteps — the old value divided by
`predict_last_n` when nothing is missing. I measured exactly 1.000 / 2.000 /
8.000 / 365.000 for `predict_last_n` = 1 / 2 / 8 / 365.

Why it's worth stating out loud: all eight CMAL configs in the repo use
`predict_last_n: 8` with `clip_gradient_norm: 1`, and on a real LSTM+CMAL head
(bs=256, T=8) the pre-clip gradient norm goes from **1.223 (clipped)** to
**0.153 (not clipped)**. So gradient clipping stops engaging, and the
optimization path changes even under Adam.

The docstring edit I have in mind is around line 378 — that the loss is averaged
over *observed* timesteps and returns 0 for an all-missing window.

If you'd rather preserve the old magnitude, the scale-preserving variant is
`return -torch.sum(result) / mask.shape[0]`. But that gives up the scale
invariance across missing fractions that #275 explicitly asked for, so my
preference is your version plus the documentation.

googlehydrology/training/loss.py
grey-nearing: **Optional: consider touching all four output tensors here, so every head parameter receives a gradient on an all-missing batch.**

The early return only touches `mu`, so `b`, `tau` and `pi` come back with
`grad is None` for an all-missing window (verified). Harmless today — we don't use
DDP, and Adam skips `None` grads — but it would trip `DistributedDataParallel`
without `find_unused_parameters=True`, and `0.0 * inf` would be NaN if the head
ever emitted a non-finite `mu`.

Costs nothing:

```suggestion
        if not torch.any(mask):
            # Touch every predicted tensor so that all head parameters
            # receive a (zero) gradient for this batch.
            zero = sum(
                prediction[key].sum() for key in ('mu', 'b', 'tau', 'pi')
            )
            return zero * 0.0
```

I applied this locally and `pytest test/test_cmal_loss_missing_values.py
test/test_losses.py` still passes (10 passed), with all four tensors getting zero
grads.

test/test_cmal_loss_missing_values.py
grey-nearing: **Optional but recommended: please add a test for the case where every sequence has a gap — that's the actual crash mode from #275, and it isn't covered yet.**

Neither existing test hits it. In the first test, row 1 is fully observed, so
`main` returns a finite (wrong) number rather than NaN, which means the NaN-abort
path — the thing that actually stops training — is currently untested.

This case fails on `main` with `isfinite(nan)` and passes here with the exact
value `log 4`. I ran it both ways:

```suggestion
    assert torch.equal(mu.grad, torch.zeros_like(mu))


@pytest.mark.unit
def test_cmal_loss_is_finite_when_every_sequence_has_a_gap():
    loss_fn = MaskedCMALLoss(_config())

    mu = torch.zeros(2, 2, 3)
    b = torch.ones(2, 2, 3)
    tau = torch.full((2, 2, 3), 0.5)
    pi = torch.full((2, 2, 3), 1.0 / 3)
    # Every sequence has one gap, so the old batch-level mask was
    # all-False and the loss came out NaN, which aborts training
    # (allow_subsequent_nan_losses defaults to 0).
    y = torch.tensor([[[0.0], [torch.nan]], [[torch.nan], [0.0]]])

    total_loss, _ = loss_fn({'mu': mu, 'b': b, 'tau': tau, 'pi': pi}, {'y': y})

    assert torch.isfinite(total_loss)
    assert torch.allclose(total_loss, torch.log(torch.tensor(4.0)), atol=1e-6)
```

googlehydrology/training/loss.py
grey-nearing: **Please add a note to the PR description and a sentence to the class docstring saying the loss is now averaged over observed timesteps. No other change needed — keep `clip_gradient_norm: 1` as it is.**

This is the line that rescales things: `main` did `-mean(sum(result, dim=1))` (sum
over timesteps, mean over batch), and this is `-mean(result)` over observed
timesteps — the old value divided by `predict_last_n` when nothing is missing. I
measured exactly 1.000 / 2.000 / 8.000 / 365.000 for `predict_last_n` = 1 / 2 / 8
/ 365.

**On the clipping threshold**, since it's the natural follow-up question and I'd
rather answer it than leave you guessing. Gradient clipping here is a guard
against exploding gradients, not a regularizer. Under your change it stops
engaging in normal training, and we think that's *correct*:

- Under the old scale, clipping was binding at initialisation (norm 1.223 vs a
  threshold of 1), which means it was acting as a step-size limiter rather than an
  explosion guard. That was probably never the intent.
- Under the new scale the norm is ~0.153, so a threshold of 1 sits roughly 6.5x
  above healthy gradients — dormant normally, still catching order-of-magnitude
  spikes. That's the shape a safety net should have.
- More importantly, a fixed threshold only becomes *meaningful* once the loss is
  per-observation. Before your change, `clip_gradient_norm: 1` meant something
  different in every config, because the scale moved with `predict_last_n` and
  with how gappy the data was. Now it doesn't.

We considered auto-scaling the threshold by `predict_last_n` and decided against
it: it would reintroduce exactly the coupling your change removes, and it wouldn't
reproduce the old behaviour on gappy data anyway, since the old scale depended on
observation count and the new one doesn't.

So: nothing to change here beyond documentation. The docstring edit I have in mind
is around line 378 — that the loss is averaged over *observed* timesteps and
returns 0 for an all-missing window.

Separately, I've opened #298 to log how often clipping actually fires, since none
of the above was visible from any training log. That's trainer-level
instrumentation and explicitly **not** something I'm asking you to take on in this
PR.

googlehydrology/training/loss.py
grey-nearing: **Optional: consider touching all four output tensors here, so every head parameter receives a gradient on an all-missing batch.**

The early return only touches `mu`, so `b`, `tau` and `pi` come back with
`grad is None` for an all-missing window (verified). Harmless today — we don't use
DDP, and Adam skips `None` grads — but it would trip `DistributedDataParallel`
without `find_unused_parameters=True`, and `0.0 * inf` would be NaN if the head
ever emitted a non-finite `mu`.

Costs nothing:

```suggestion
        if not torch.any(mask):
            # Touch every predicted tensor so that all head parameters
            # receive a (zero) gradient for this batch.
            zero = sum(
                prediction[key].sum() for key in ('mu', 'b', 'tau', 'pi')
            )
            return zero * 0.0
```

I applied this locally and `pytest test/test_cmal_loss_missing_values.py
test/test_losses.py` still passes (10 passed), with all four tensors getting zero
grads.

test/test_cmal_loss_missing_values.py
grey-nearing: **Optional but recommended: please add a test for the case where every sequence has a gap — that's the actual crash mode from #275, and it isn't covered yet.**

Neither existing test hits it. In the first test, row 1 is fully observed, so
`main` returns a finite (wrong) number rather than NaN, which means the NaN-abort
path — the thing that actually stops training — is currently untested.

This case fails on `main` with `isfinite(nan)` and passes here with the exact
value `log 4`. I ran it both ways:

```suggestion
    assert torch.equal(mu.grad, torch.zeros_like(mu))


@pytest.mark.unit
def test_cmal_loss_is_finite_when_every_sequence_has_a_gap():
    loss_fn = MaskedCMALLoss(_config())

    mu = torch.zeros(2, 2, 3)
    b = torch.ones(2, 2, 3)
    tau = torch.full((2, 2, 3), 0.5)
    pi = torch.full((2, 2, 3), 1.0 / 3)
    # Every sequence has one gap, so the old batch-level mask was
    # all-False and the loss came out NaN, which aborts training
    # (allow_subsequent_nan_losses defaults to 0).
    y = torch.tensor([[[0.0], [torch.nan]], [[torch.nan], [0.0]]])

    total_loss, _ = loss_fn({'mu': mu, 'b': b, 'tau': tau, 'pi': pi}, {'y': y})

    assert torch.isfinite(total_loss)
    assert torch.allclose(total_loss, torch.log(torch.tensor(4.0)), atol=1e-6)
```

googlehydrology/training/loss.py
grey-nearing, line 427: **Please add a sentence to the class docstring, and to the PR description, saying the loss is averaged over observed timesteps. Nothing else to change here — `clip_gradient_norm: 1` can stay as it is.**

This is the line that rescales things: `main` did `-mean(sum(result, dim=1))` (sum
over timesteps, mean over batch), and this is `-mean(result)` over observed
timesteps. The ratio is exactly `predict_last_n` when nothing is missing — I
measured 1.000 / 2.000 / 8.000 / 365.000 for `predict_last_n` = 1 / 2 / 8 / 365.

The docstring line I have in mind is around line 378: that the loss is averaged
over *observed* timesteps, and returns 0 for an all-missing window.

test/test_cmal_loss_missing_values.py
grey-nearing, line 69: **Optional but recommended: please add a test for the case where every sequence has a gap — that's the actual crash mode from #275, and it isn't covered yet.**

Neither existing test hits it. In the first test, row 1 is fully observed, so
`main` returns a finite (wrong) number rather than NaN, which means the NaN-abort
path — the thing that actually stops training — is currently untested.

This case fails on `main` with `isfinite(nan)` and passes here with the exact
value `log 4`. I ran it both ways:

```suggestion
    assert torch.equal(mu.grad, torch.zeros_like(mu))


@pytest.mark.unit
def test_cmal_loss_is_finite_when_every_sequence_has_a_gap():
    loss_fn = MaskedCMALLoss(_config())

    mu = torch.zeros(2, 2, 3)
    b = torch.ones(2, 2, 3)
    tau = torch.full((2, 2, 3), 0.5)
    pi = torch.full((2, 2, 3), 1.0 / 3)
    # Every sequence has one gap, so the old batch-level mask was
    # all-False and the loss came out NaN, which aborts training
    # (allow_subsequent_nan_losses defaults to 0).
    y = torch.tensor([[[0.0], [torch.nan]], [[torch.nan], [0.0]]])

    total_loss, _ = loss_fn({'mu': mu, 'b': b, 'tau': tau, 'pi': pi}, {'y': y})

    assert torch.isfinite(total_loss)
    assert torch.allclose(total_loss, torch.log(torch.tensor(4.0)), atol=1e-6)
```

grey-nearing (commented):
AI review approved by me.

sylvesterkaczmarek:
Addressed in 16285c7c. The class docstring and PR description now state that CMAL loss is averaged over observed timesteps, and I added the all-sequences-have-a-gap regression using the reviewer’s suggested shape. git diff --check and Python syntax compilation pass locally; pytest is not installed in this local checkout, so I left runtime validation to the hosted CI.

grey-nearing (approved):
Approving — both requests are addressed and CI is green. Thanks for the fix!

PR #291, opened 2026-09-10
Apply negative clipping to deterministic CMAL forecasts

### LINKED ISSUE #277: [Bug] sample_cmal_deterministic ignores negative_sample_handling config and produces negative streamflow values
### Description
  When a model is configured with `negative_sample_handling: 'clip'`, predictions are expected to enforce physical non-negativity ($Q \ge 0.0$).

  While `sample_cmal()` correctly ingests the dataset `Scaler`, calculates the normalized zero threshold, and applies `_handle_negative_values()`, `sample_cmal_deterministic()` omits the `scaler` parameter and bypasses negative sample handling entirely. Consequently, deterministic CMAL forecasts    
produce negative streamflow predictions when unscaled.

  In `test/test_uncertainty.py:137-144`, this discrepancy was bypassed by explicitly exempting `cmal_deterministic` from the non-negativity check:

  ```python
  if (
      negative_sample_handling == 'clip'
      and config.head.lower() != 'cmal_deterministic'
  ):
      assert np.allclose(negative_vals, 0.0, atol=1e-6)
```
### Steps to Reproduce

1. Configure a run with head: 'cmal_deterministic' and negative_sample_handling: 'clip'.
2. Generate predictions during low-flow periods where lower quantiles (q = 0.1,q = 0.2) fall below normalized zero (-center/scale).
3. Unscale predictions to physical space (m³/s).
4. Observed: Lower quantiles evaluate to negative streamflow discharge.
5. Expected: Values below normalized zero are clamped, producing non-negative streamflow (≥0.0).

### Affected Files

• googlehydrology/utils/samplingutils.py (lines 378-415)
• test/test_uncertainty.py (lines 137-144)

### Proposed Fix

1. Update sample_cmal_deterministic in googlehydrology/utils/samplingutils.py to accept scaler: Scaler | None = None and execute _handle_negative_values():
  ```python
  def sample_cmal_deterministic(
      model: 'BaseModel',
      data: dict[str, torch.Tensor],
      *,
      scaler: Scaler | None = None,
      outputs: dict[str, torch.Tensor] | None = None,
  ) -> dict[str, torch.Tensor]:
      # ... generate sample points ...
      if scaler is not None and setup.cfg.negative_sample_handling:
          normalized_zero = _calc_normalized_zero_thresholds(
              scaler=scaler,
              targets=setup.cfg.target_variables,
              device=setup.device,
              dtype=samples[f'y_hat{setup.freq_suffixes[0]}'].dtype,
          )
          for freq_suffix in setup.freq_suffixes:
              key = f'y_hat{freq_suffix}'
              samples[key] = _handle_negative_values(
                  cfg=setup.cfg,
                  values=samples[key],
                  sample_values=lambda ids: samples[key][ids],
                  normalized_zero=normalized_zero,
              )
      return samples
```
2. Pass scaler through sample_pointpredictions() at samplingutils.py:65.
3. Remove the and config.head.lower() != 'cmal_deterministic' exception from test/test_uncertainty.py:139.

sylvesterkaczmarek:
Fixes #277.

Passes the run scaler into deterministic CMAL sampling and applies normalized-zero clipping to deterministic summary outputs when negative_sample_handling is clip. Unsupported handling values now follow the shared negative-value handler and raise instead of being silently ignored; truncate remains a documented no-op because deterministic CMAL emits summary statistics rather than redrawable samples.

Breaking API note: scaler is now a required positional argument to sample_cmal_deterministic. The only in-tree caller has been updated, but direct external/notebook callers need to pass the run scaler.

The mixture mean is clipped as a physical floor. This is max(E[X], 0), not a distributional correction equivalent to E[max(X, 0)].

Adds focused regression coverage for clip, none, unset, truncate and unsupported handling modes, and removes the integration-test exemption for deterministic CMAL.

Validation:
- Python syntax compilation
- git diff --check
- hosted cross-platform PyTest CI will validate the updated head

### DIFF
--- a/googlehydrology/utils/samplingutils.py
+++ b/googlehydrology/utils/samplingutils.py
@@ -62,7 +62,9 @@ def sample_pointpredictions(
     if model.cfg.head.lower() == 'cmal':
         samples = sample_cmal(model, data, n_samples, scaler, outputs=outputs)
     elif model.cfg.head.lower() == 'cmal_deterministic':
-        samples = sample_cmal_deterministic(model, data, outputs=outputs)
+        samples = sample_cmal_deterministic(
+            model, data, scaler, outputs=outputs
+        )
     elif model.cfg.head.lower() == 'regression':  # regression head assumes mcd
         assert not outputs, 'regression self creates outputs'
         samples = sample_mcd(model, data, n_samples, scaler)
@@ -330,6 +332,7 @@ def sample_mcd(
 def sample_cmal_deterministic(
     model: 'BaseModel',
     data: dict[str, torch.Tensor],
+    scaler: Scaler,
     *,
     outputs: dict[str, torch.Tensor] | None = None,
 ) -> dict[str, torch.Tensor]:
@@ -344,6 +347,8 @@ def sample_cmal_deterministic(
         A model with a CMAL head.
     data : dict[str, torch.Tensor]
         Dictionary, containing input features as key-value pairs.
+    scaler : Scaler
+        Scaler of the run.
     outputs, optional
         Model forward result
 
@@ -364,6 +369,15 @@ def sample_cmal_deterministic(
     # not point predictions.
     pred = outputs or model(data)
 
+    normalized_zeros = None
+    if (setup.cfg.negative_sample_handling or '').lower() == 'clip':
+        normalized_zeros = _calc_normalized_zero_thresholds(
+            scaler=scaler,
+            targets=setup.cfg.target_variables,
+            device=next(model.parameters()).device,
+            dtype=next(model.parameters()).dtype,
+        )
+
     # Map output frequencies to final sample tensors:
     samples = {}
 
@@ -374,10 +388,15 @@ def sample_cmal_deterministic(
         tau = pred[f'tau{freq_suffix}']  # asymmetries
         pi = pred[f'pi{freq_suffix}']  # weights
 
-        sample_points = [
-            cmal_deterministic.generate_predictions(mu, b, tau, pi)
-        ]
-        samples[f'y_hat{freq_suffix}'] = torch.stack(sample_points, 2)
+        values = cmal_deterministic.generate_predictions(mu, b, tau, pi)
+        if normalized_zeros is not None:
+            values = _handle_negative_values(
+                setup.cfg,
+                values,
+                sample_values=lambda ids: values[ids],
+                normalized_zero=normalized_zeros[0],
+            )
+        samples[f'y_hat{freq_suffix}'] = torch.stack([values], 2)
 
     return samples
 
@@ -569,4 +588,3 @@ def sample_cmal(
         samples[f'y_hat{freq_suffix}'] = torch.stack(sample_points, dim=2)
 
     return samples
-
--- a/test/test_cmal_deterministic_clipping.py
+++ b/test/test_cmal_deterministic_clipping.py
@@ -1,0 +1,70 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from unittest.mock import MagicMock
+
+import pytest
+import torch
+import xarray as xr
+
+from googlehydrology.utils import samplingutils
+
+
+@pytest.mark.unit
+def test_deterministic_cmal_clips_at_normalized_zero(monkeypatch):
+    model = MagicMock()
+    model.parameters.side_effect = lambda: iter([torch.zeros(1)])
+    model.cfg.head = 'cmal_deterministic'
+    model.cfg.mc_dropout = False
+    model.cfg.target_variables = ['streamflow']
+    model.cfg.use_frequencies = ['1D']
+    model.cfg.predict_last_n = {'1D': 2}
+    model.cfg.negative_sample_handling = 'clip'
+
+    scaler = MagicMock()
+    scaler.scaler = {
+        'streamflow': xr.DataArray(
+            [5.0, 2.0],
+            coords={'parameter': ['center', 'scale']},
+            dims=['parameter'],
+        )
+    }
+
+    generated = torch.tensor(
+        [[[-3.0, -2.0, -1.0, 0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0],
+          [-4.0, -2.5, -1.5, 0.5, 1.5, 2.5, 3.5, 4.5, 5.5, 6.5]]]
+    )
+    monkeypatch.setattr(
+        samplingutils.cmal_deterministic,
+        'generate_predictions',
+        lambda *args: generated.clone(),
+    )
+
+    outputs = {
+        'mu': torch.zeros(1, 2, 3),
+        'b': torch.ones(1, 2, 3),
+        'tau': torch.full((1, 2, 3), 0.5),
+        'pi': torch.full((1, 2, 3), 1.0 / 3),
+    }
+    samples = samplingutils.sample_pointpredictions(
+        model,
+        {'y': torch.zeros(1, 2, 1)},
+        n_samples=10,
+        scaler=scaler,
+        outputs=outputs,
+    )
+
+    assert samples['y_hat'].shape == (1, 2, 1, 10)
+    assert samples['y_hat'].min().item() == -2.5
+    assert samples['y_hat'][0, 0, 0, 1].item() == -2.0
--- a/test/test_uncertainty.py
+++ b/test/test_uncertainty.py
@@ -50,7 +50,6 @@ def test_daily_uncertainty(
     ('none', 'clip', 'truncate') and with or without Monte Carlo dropout.
     """
 
-    # TODO (future) :: clip fails with CMAL-deterministic, should call _handle_negative_values
     config = get_config('forecast')
     updates = {
         'model': forecast_model,
@@ -134,10 +133,7 @@ def _check_uncertainty_output(
     negative_vals = test_vals[sample_key].values[
         test_vals[sample_key].values < 0
     ]
-    if (
-        negative_sample_handling == 'clip'
-        and config.head.lower() != 'cmal_deterministic'
-    ):
+    if negative_sample_handling == 'clip':
         # For 'clip', we expect all non-negative values
         min_val = np.min(negative_vals) if len(negative_vals) > 0 else 0.0
         assert np.allclose(negative_vals, 0.0, atol=1e-6), (


googlehydrology/utils/samplingutils.py
```diff
@@ -330,6 +332,7 @@ def _sample_values(ids: list[int]) -> torch.Tensor:
 def sample_cmal_deterministic(
     model: 'BaseModel',
     data: dict[str, torch.Tensor],
+    scaler: Scaler,
```
grey-nearing: 🟢 **Non-blocking, API note.** Making `scaler` a required positional arg is the right call in my view — it matches `sample_cmal(model, data, n_samples, scaler, ...)` and it makes it impossible to silently skip clipping again (which is exactly how #277 happened). Note this is a hard break for anyone calling `sample_cmal_deterministic(model, data, outputs=...)` directly from a notebook; I grepped the repo and the only in-tree caller is `sample_pointpredictions`, so I am fine taking the break rather than the `scaler: Scaler | None = None` variant proposed in the issue. Please just mention it in the PR description so it shows up in the release notes.

googlehydrology/utils/samplingutils.py
```diff
@@ -364,6 +369,15 @@ def sample_cmal_deterministic(
...
     pred = outputs or model(data)
 
+    normalized_zeros = None
+    if (setup.cfg.negative_sample_handling or '').lower() == 'clip':
+        normalized_zeros = _calc_normalized_zero_thresholds(
+            scaler=scaler,
+            targets=setup.cfg.target_variables,
+            device=next(model.parameters()).device,
+            dtype=next(model.parameters()).dtype,
+        )
```
grey-nearing: 🟡 **The `== 'clip'` guard makes this head silently swallow invalid config values.** For the `cmal` head, `negative_sample_handling: 'bogus'` raises `NotImplementedError` out of `_handle_negative_values`. With this guard, the same typo on a `cmal_deterministic` run is silently ignored and you get unclipped negative discharge. I verified both branches on your head commit:

| `negative_sample_handling` | `cmal` | `cmal_deterministic` (this PR) |
|---|---|---|
| `'clip'` | clipped | clipped ✅ |
| `'truncate'` | resampled | **ignored** |
| `'bogus'` | `NotImplementedError` | **silently ignored** |

Letting everything except the modes that genuinely do not apply fall through to the shared helper fixes the error surface and documents why `truncate` is a no-op for a head that emits summary statistics rather than draws. I applied this exact change locally: `test/test_cmal_deterministic_clipping.py` + `test/test_samplingutils.py` (9 passed) and all 6 `test_uncertainty.py::test_daily_uncertainty[...cmal_deterministic...]` cases pass, and `'bogus'` now raises.

```suggestion
    # This head returns summary statistics (the mixture mean plus nine
    # quantiles), not random draws, so 'truncate' has nothing to resample and
    # is a no-op here. Everything else is delegated to
    # `_handle_negative_values`, so an unsupported value still raises.
    negative_handling = (setup.cfg.negative_sample_handling or '').lower()
    normalized_zeros = None
    if negative_handling not in ('', 'none', 'truncate'):
        normalized_zeros = _calc_normalized_zero_thresholds(
            scaler=scaler,
            targets=setup.cfg.target_variables,
            device=setup.device,
            dtype=next(model.parameters()).dtype,
        )
```

googlehydrology/utils/samplingutils.py
```diff
@@ -374,10 +388,15 @@ def sample_cmal_deterministic(
         tau = pred[f'tau{freq_suffix}']  # asymmetries
         pi = pred[f'pi{freq_suffix}']  # weights
 
-        sample_points = [
-            cmal_deterministic.generate_predictions(mu, b, tau, pi)
-        ]
-        samples[f'y_hat{freq_suffix}'] = torch.stack(sample_points, 2)
+        values = cmal_deterministic.generate_predictions(mu, b, tau, pi)
```
grey-nearing: 🟡 **Statistical caveat worth a comment in the code: element 0 of `generate_predictions` is the mixture *mean*, not a quantile, and clipping a mean is not the same operation as clipping a sample.**

For the nine quantiles this PR is exactly right: the quantile function is monotone, so `max(Q_X(p), z) == Q_max(X,z)(p)`. I checked numerically against 2e6 draws from the same mixture and the clipped deterministic quantiles match the empirical quantiles of the clipped samples to 0.0000 on every clipped quantile — i.e. the deterministic path is now *consistent* with the stochastic path. 👍

Element 0 is the odd one out. Clipping gives `max(E[X], z)`, whereas the stochastic path with `clip` gives `E[max(X, z)]`, and those differ by Jensen. With a `center=2.0, scale=4.0` streamflow scaler (`z = -0.5`) on three realistic mixtures I get:

| case | `max(E[X], z)` (this PR) | `E[max(X, z)]` (stochastic `clip`) | bias |
|---|---|---|---|
| low flow | -0.338 | -0.053 | **-1.14 m³/s** |
| very low flow | -0.500 | -0.293 | **-0.83 m³/s** |
| mid flow | 0.515 | 0.753 | **-0.95 m³/s** |

Two things follow. (1) This PR still strictly improves the "very low flow" case: the mean element goes from -0.277 m³/s (physically impossible) to exactly 0.0. (2) The residual gap is pre-existing — the mean element has always been `E[X]` rather than the clipped mean — so I am **not** asking you to fix it here. I would just like a one-line comment noting that element 0 is a mean and that clipping it is a floor, not a distributional correction, so the next reader does not assume this path is moment-consistent with `sample_cmal`.

If you (or a follow-up) ever want the consistent version, the ALD mixture has it in closed form; I verified this against 4e6 Monte Carlo draws (agreement ≤ 5e-4, i.e. MC noise):

```python
# E[max(X, z)] for one ALD component, repo parameterization
tau_c = 1 - tau
above = z + (b * tau_c / tau) * torch.exp(-tau * (z - mu) / b)
below = (
    mu
    - (tau * b / tau_c) * (1 - torch.exp((z - mu) * tau_c / b))
    + b * tau_c / tau
)
clipped_mean = (pi * torch.where(z > mu, above, below)).sum(-1)
```

googlehydrology/utils/samplingutils.py
```diff
@@ -374,10 +388,15 @@ def sample_cmal_deterministic(
...
+        if normalized_zeros is not None:
+            values = _handle_negative_values(
+                setup.cfg,
+                values,
+                sample_values=lambda ids: values[ids],
+                normalized_zero=normalized_zeros[0],
+            )
```
grey-nearing: 🟢 **Two latent traps here, both currently harmless — please just document them.**

1. `sample_values=lambda ids: values[ids]` is dead code today (only `clip` reaches it), but it is a no-op resampler: if the guard above ever loosens to include `'truncate'`, the retry loop in `_handle_negative_values` would assign `values[mask] = values[mask]` and spin `negative_sample_max_retries` times changing nothing, then return the negatives anyway. Silent wrong answer rather than a loud failure.
2. `normalized_zeros[0]` hardcodes target 0 while `_calc_normalized_zero_thresholds` returns one threshold per target. I confirmed this is fine *today* only because `generate_predictions` collapses all targets into one mixture — with `target_variables = [streamflow, tmax]` the returned tensor is still `[1, 2, 1, 10]` (target dim 1) on both `main` and this branch, so multi-target `cmal_deterministic` is already broken upstream of your change. But if someone fixes that, this line will quietly clip `tmax` at the streamflow threshold.

Comment-only suggestion (verified: 9 unit + 6 integration tests still pass with it applied):

```suggestion
        if normalized_zeros is not None:
            values = _handle_negative_values(
                setup.cfg,
                values,
                # Unused: 'clip' never resamples, and a summary statistic
                # cannot be redrawn.
                sample_values=lambda _: values,
                # `generate_predictions` collapses every target into a single
                # mixture, so `values` only ever holds target 0.
                normalized_zero=normalized_zeros[0],
            )
```

test/test_cmal_deterministic_clipping.py
```diff
@@ -0,0 +1,70 @@
...
+            [5.0, 2.0],
+            coords={'parameter': ['center', 'scale']},
+            dims=['parameter'],
+        )
+    }
+
+    generated = torch.tensor(
+        [[[-3.0, -2.0, -1.0, 0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0],
+          [-4.0, -2.5, -1.5, 0.5, 1.5, 2.5, 3.5, 4.5, 5.5, 6.5]]]
+    )
```
grey-nearing: 🟢 Nit, `ruff format` (not in CI, so non-blocking) wants this block reflowed:

```suggestion
    generated = torch.tensor(
        [
            [
                [-3.0, -2.0, -1.0, 0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0],
                [-4.0, -2.5, -1.5, 0.5, 1.5, 2.5, 3.5, 4.5, 5.5, 6.5],
            ]
        ]
    )
```

Separately — and this is the more useful suggestion — the test would be stronger as a parametrized pair with a negative control, so a future refactor that clips unconditionally also gets caught:

```python
@pytest.mark.parametrize('handling,expected_min', [('clip', -2.5), ('none', -4.0), (None, -4.0)])
def test_deterministic_cmal_negative_handling(
    monkeypatch, handling, expected_min
):
    ...
    assert samples['y_hat'].min().item() == expected_min
```

I ran exactly this shape against your branch: `'none'`/`None` give -4.0 and `clip` gives -2.5, so the control is meaningful and cheap.

test/test_uncertainty.py
```diff
@@ -134,10 +133,7 @@ def _check_uncertainty_output(
     negative_vals = test_vals[sample_key].values[
         test_vals[sample_key].values < 0
     ]
-    if (
-        negative_sample_handling == 'clip'
-        and config.head.lower() != 'cmal_deterministic'
-    ):
+    if negative_sample_handling == 'clip':
```
grey-nearing: 🟢 **This is the part of the PR I care most about, and it holds up.** I re-ran the integration test with your test changes but `samplingutils.py` reverted to `main`:

```
FAILED test_daily_uncertainty[...-cmal_deterministic-clip-False]
FAILED test_daily_uncertainty[...-cmal_deterministic-clip-True]
AssertionError: Found negative samples below tolerance. Smallest val: -2.368692398071289
```

and with your fix both pass (25 s). So removing the exemption is a genuine regression guard, not a cosmetic cleanup, and the -2.37 m³/s figure confirms #277 was real in an end-to-end run. Thank you for deleting the TODO rather than leaving it dangling.

googlehydrology/utils/samplingutils.py
grey-nearing, line 335: **No code change needed — please just mention this breaking change in the PR description so it reaches the release notes.**

Making `scaler` a required positional argument is the right call. It matches
`sample_cmal(model, data, n_samples, scaler, ...)`, and it makes it impossible to
silently skip clipping again, which is exactly how #277 happened.

It is a hard break for anyone calling `sample_cmal_deterministic(model, data,
outputs=...)` directly from a notebook. I grepped the repo and the only in-tree
caller is `sample_pointpredictions`, so I'd rather take the clean break than the
`scaler: Scaler | None = None` variant.

googlehydrology/utils/samplingutils.py
grey-nearing, line 384: **Please change this guard so that a misspelled config value raises an error instead of being silently ignored.**

This is the one functional change I'm asking for. Right now a typo in
`negative_sample_handling` behaves differently depending on which head you use.
I verified both branches on your head commit:

| `negative_sample_handling` | `cmal` | `cmal_deterministic` (this PR) |
|---|---|---|
| `'clip'` | clipped | clipped ✅ |
| `'truncate'` | resampled | **ignored** |
| `'bogus'` | `NotImplementedError` | **silently ignored** |

So a user who typos this on a deterministic run gets unclipped negative discharge
and no warning.

Letting everything through to the shared helper *except* the modes that genuinely
don't apply fixes the error surface, and documents why `truncate` is a no-op for a
head that emits summary statistics rather than draws.

I applied this exact change locally: `test_cmal_deterministic_clipping.py` +
`test_samplingutils.py` (9 passed), all 6
`test_uncertainty.py::test_daily_uncertainty[...cmal_deterministic...]` cases pass,
and `'bogus'` now raises.

```suggestion
    # This head returns summary statistics (the mixture mean plus nine
    # quantiles), not random draws, so 'truncate' has nothing to resample and
    # is a no-op here. Everything else is delegated to
    # `_handle_negative_values`, so an unsupported value still raises.
    negative_handling = (setup.cfg.negative_sample_handling or '').lower()
    normalized_zeros = None
    if negative_handling not in ('', 'none', 'truncate'):
        normalized_zeros = _calc_normalized_zero_thresholds(
            scaler=scaler,
            targets=setup.cfg.target_variables,
            device=setup.device,
            dtype=next(model.parameters()).dtype,
        )
```

googlehydrology/utils/samplingutils.py
grey-nearing, line 396: **Please add a one-line code comment noting that element 0 is the mixture mean, so a later reader doesn't assume this path is moment-consistent.**

No behaviour change requested. The nine quantiles are handled exactly right here.
Element 0 is the odd one out: clipping it gives `max(E[X], z)`, whereas the
stochastic path gives `E[max(X, z)]`, and those differ by Jensen. On three
realistic mixtures with a `center=2.0, scale=4.0` scaler (`z = -0.5`) the gap is a
0.8–1.1 m³/s low bias.

Your change still strictly improves this case — in deep low flow the mean element
goes from −0.277 m³/s, which is physically impossible, to exactly 0.0. The
residual gap is pre-existing and not yours to fix. I'd just like the next person
reading this to know that clipping a mean is a floor, not a distributional
correction.

googlehydrology/utils/samplingutils.py
grey-nearing, line 409: **No behaviour change needed — please just take the two clarifying comments in the suggestion below.**

Two things here are harmless today but would bite later:

1. `sample_values=lambda ids: values[ids]` is dead code, since only `clip` reaches
   it. But it's a no-op resampler: if the guard above ever loosens to include
   `'truncate'`, the retry loop in `_handle_negative_values` would assign
   `values[mask] = values[mask]`, spin `negative_sample_max_retries` times changing
   nothing, and return the negatives anyway — a silent wrong answer rather than a
   loud failure.
2. `normalized_zeros[0]` hardcodes target 0, while `_calc_normalized_zero_thresholds`
   returns one threshold per target. This is fine *today* only because
   `generate_predictions` collapses all targets into a single mixture. I confirmed
   that with `target_variables = [streamflow, tmax]` the returned tensor is still
   `[1, 2, 1, 10]` on both `main` and this branch, so multi-target
   `cmal_deterministic` is already broken upstream of your change. But if someone
   fixes that, this line will quietly clip `tmax` at the streamflow threshold.

Verified: 9 unit + 6 integration tests still pass with this applied.

```suggestion
        if normalized_zeros is not None:
            values = _handle_negative_values(
                setup.cfg,
                values,
                # Unused: 'clip' never resamples, and a summary statistic
                # cannot be redrawn.
                sample_values=lambda _: values,
                # `generate_predictions` collapses every target into a single
                # mixture, so `values` only ever holds target 0.
                normalized_zero=normalized_zeros[0],
            )
```

test/test_cmal_deterministic_clipping.py
grey-nearing, line 56: **Optional: this test would catch more if it had a negative control.**

As written it proves clipping happens when it's switched on, but it wouldn't catch
a future refactor that clips *unconditionally*. Parametrizing with the disabled
cases closes that gap cheaply:

```python
@pytest.mark.parametrize(
    'handling,expected_min', [('clip', -2.5), ('none', -4.0), (None, -4.0)]
)
def test_deterministic_cmal_negative_handling(monkeypatch, handling, expected_min):
    ...
    assert samples['y_hat'].min().item() == expected_min
```

I ran this shape against your branch: `'none'` and `None` both give −4.0 while
`'clip'` gives −2.5, so the control is meaningful. Entirely up to you.

test/test_uncertainty.py
grey-nearing, line 136: **Nothing to change — flagging that this is the most valuable part of the PR.**

Removing this exemption is a genuine regression guard rather than a cosmetic
cleanup. With your test changes but `samplingutils.py` reverted to `main`:

```
FAILED test_daily_uncertainty[...-cmal_deterministic-clip-False]
FAILED test_daily_uncertainty[...-cmal_deterministic-clip-True]
AssertionError: Found negative samples below tolerance. Smallest val: -2.368692398071289
```

Both pass with your fix (25 s). Thank you for deleting the TODO rather than
leaving it dangling.

grey-nearing (commented):
AI review approved by me.

sylvesterkaczmarek:
Addressed in signed commit 7fb3f45a. Deterministic CMAL now treats truncate as an explicit no-op, delegates every other non-empty/non-none mode to the shared negative-value handler so typos raise, and documents the mixture-mean clipping caveat plus the target-0 assumption. I also added negative controls for none/unset/truncate and an unsupported-mode regression. The PR description now calls out the required-scaler API break. git diff --check and Python syntax compilation pass locally; hosted CI will provide the runtime matrix.

grey-nearing (approved):
Approving — all review comments are addressed and CI is green. Thanks for the fix!

sylvesterkaczmarek:
Thanks for the detailed reviews and for merging both CMAL fixes. Much appreciated.

PR #293, opened 2026-09-16
Catchment Delineation Module

grey-nearing:
## Summary

Adds the `catchment_delineation` package (`delineate-catchment` and `benchmark-catchment` CLIs) for tracing upstream watershed boundaries (polygons) and drainage areas ($\text{km}^2$) from 90-meter (3-arc-second) D8 flow-direction maps.

## Key Features

- **Upstream Flow Tracing Across Tile Borders**: Traces upstream cells across 5°×5° (`6000 × 6000`) flow-direction tiles using a C-accelerated + 1D-vectorized BFS engine and stitches multi-tile basins into a single continuous polygon (`max_cells=None` by default so large rivers are never cut off).
- **Stream Channel Snapping + Optional Expected Area Hint (`--expected-area` / `--area-col`)**: Snaps gauge coordinates onto the nearest stream channel (`12` cells ≈ `1.1 km` by default). When an approximate expected drainage area ($\text{km}^2$) is provided, searches up to `80` cells (`~7.2 km`) for the river channel matching that expected area (within `±50%` by default). If no nearby channel matches the expected area, logs a `[AREA HINT FAILURE]` error and refuses to output a polygon (`CatchmentAreaMismatchError`).
- **Explicit User-Supplied Paths & Loud Failure on Missing Data**: All input and output paths (`--tiles-dir`, `--gcs-uri` + `--cache-dir`, `-o` / `--output-dir`) are supplied explicitly by the user with no hardcoded buckets or fallback directories. If a required tile is missing or a watershed crosses outside the DEM coverage (`-56°S` to `60°N`), raises `CatchmentCoverageError` instead of producing a truncated or zero-filled polygon.
- **Caravan Folder Layout Support**: `--preserve-caravan-dirs` saves outputs under `<output-dir>/shapefiles/<subdataset>/<subdataset>_basin_shapes.{geoparquet,geojson,shp}`.

## 1,200-Basin Benchmark Results

Evaluated across `1,200` global gauges (`1,127` within the `[-56°S, 60°N]` HydroSHEDS coverage bounds, `73` high-latitude gauges above `60°N` explicitly flagged out-of-coverage):

| Basin Size Bucket | In-Coverage Basins | Median IoU | Median Dice | Basins with $\text{IoU} \ge 0.80$ | Median Area Error |
| :--- | :---: | :---: | :---: | :---: | :---: |
| **1_micro** ($< 10\text{ km}^2$) | 132 | **0.916** | **0.956** | **87.1%** (115 / 132) | 5.29% |
| **2_small** ($10\text{–}100\text{ km}^2$) | 172 | **0.955** | **0.977** | **98.3%** (169 / 172) | 2.68% |
| **3_medium** ($100\text{–}1,000\text{ km}^2$) | 222 | **0.975** | **0.988** | **98.2%** (218 / 222) | 1.88% |
| **4_large** ($1,000\text{–}10,000\text{ km}^2$) | 282 | **0.984** | **0.992** | **98.9%** (279 / 282) | 1.38% |
| **5_macro** ($> 10,000\text{ km}^2$) | 319 | **0.991** | **0.995** | **99.4%** (317 / 319) | 0.96% |
| **All In-Coverage** | **1,127** | **0.976** | **0.988** | **97.4%** (1,098 / 1,127) | **1.80%** |

## Usage

```bash
# Single coordinate pair
delineate-catchment \
  --lat 39.6828 --lon -88.7729 \
  --expected-area 480.0 \
  --tiles-dir /path/to/tiles_5deg \
  -o basin.geojson --pretty

# Batch from CSV with Caravan directory layout
delineate-catchment \
  --csv gauges.csv \
  --area-col expected_area_km2 \
  --tiles-dir /path/to/tiles_5deg \
  --output-dir /path/to/caravan_output \
  --preserve-caravan-dirs \
  --format all --workers 8
```


grey-nearing:
## Terminology note: this is a **benchmark**, and that's the right word

Cross-posting to #292, #293 and #294 so we have one shared definition, since "canary" and "benchmark" have been getting used interchangeably in discussion.

Having read all three PRs: **no change is needed here.** This PR is correctly called a benchmark, and the term is used consistently — "canary" appears zero times in #292 and #293, and "benchmark" appears zero times in #294. The ambiguity is in how we *talk* about them, not in the code. This comment is just to keep it that way.

### The distinction

| | **Benchmark** | **Canary** |
| --- | --- | --- |
| Question | "Is our output *accurate*?" | "Is the upstream feed *still working*?" |
| Compares against | A frozen ground-truth reference dataset | Nothing — checks data arrives and is plausible |
| Output | Scores (IoU, Dice, Pearson, Spearman) | Pass / fail |
| Fails when | **You** regress something | **A third party** changes or breaks something |
| Determinism | Deterministic | Non-deterministic (live network) |
| Cadence | On demand, pre-release | Scheduled (nightly) |
| Gates a PR? | May gate a release | **Never** |

The rule of thumb that makes it stick:

> **A benchmark fails when *you* break something. A canary fails when *someone else* does.**

That asymmetry is the whole reason they must be wired up differently — a canary that gates merges means a third-party outage blocks the team.

### Why each PR landed on the word it did

- **#292** scores extracted static attributes against published Caravan reference values across ~490 basins → correlation metrics. **Benchmark.**
- **#293** scores delineated catchments against reference polygons across ~1,200 basins → IoU / Dice / area error. **Benchmark.**
- **#294** is ETL: it copies and reshapes upstream data. There is no ground truth to score against, so "benchmark" doesn't apply. What it needs is liveness detection for three fragile external feeds (NOAA PSL URLs, the WeatherBench 2 Zarr, ECMWF's GRIB index layout). **Canary.**

One thing worth being explicit about: the benchmarks in #292/#293 *do* reach the network — they pull reference data from `gs://open-multimet/ancillary-data/benchmarks/`. That makes them look canary-ish operationally, but they're still benchmarks; the GCS fetch is just how the frozen reference data is delivered. The reference data doesn't change under you, so a failure still means *you* regressed.

### Proposed shared vocabulary

Four categories, now written up in the top-level `README.md` in #294:

| Kind | Question | Data source | Gates a PR? |
| --- | --- | --- | --- |
| **Test** | Is the code correct? | Synthetic, hermetic | **Yes** |
| **Benchmark** | Is the output accurate? | Frozen reference dataset | No — on demand |
| **Canary** | Is the upstream feed working? | Live third-party services | **Never** |
| **Audit** | Is a *published archive* intact? | The archive in GCS | No — on demand |

The fourth one, **audit**, is worth naming separately because we've been doing it ad hoc — e.g. the recent IMERG repair, where a published Zarr store had 36 duplicate timestamps from an interrupted run. That's not a test, a benchmark, or a canary: the code was fine, the upstream was fine, and the reference data was irrelevant. The *artifact* was corrupt. Conflating that with the other three is how it went unnoticed for as long as it did.

No action needed on this PR.

environments/environment_cpu.yml
```diff
@@ -21,40 +21,45 @@ channels:
...
-  - torchaudio=2.5.*
-  - torchvision=0.20.*
-  - absl-py=2.3.*
-  - bokeh=3.8.*
-  - cachey=0.2.*
-  - dask=2025.12.*
-  - gcsfs=2025.12.*
-  - geopandas=1.1.*
+  - torchaudio
+  - torchvision
```
grey-nearing: 🔴 **Blocking — please restore the version pins.**

This file (and `conda.yml`) has had *every* pin removed. `environment_cpu.yml`
is what CI installs, so from here on CI resolves whatever torch/numpy/xarray
happen to be newest that day, and a green run today says nothing about
tomorrow. We ship pretrained checkpoints and retrain rarely — an unpinned
`pytorch` means a checkpoint saved this week may not load next month, and
nobody will connect the failure to this PR.

The full delta vs `main`: `pytorch 2.5.*`, `torchvision 0.20.*`,
`torchaudio 2.5.*`, `numpy 2.4.*`, `pandas 2.3.*`, `xarray 2025.12.*`,
`zarr 3.1.*`, `dask`, `gcsfs`, `geopandas`, `scipy`, `sphinx`, `tensorboard`,
`bokeh`, `absl-py`, `h5py`, `h5netcdf`, `netcdf4`, `matplotlib`,
`more-itertools`, `nbsphinx`, `parso`, `pydantic`, `ruamel.yaml` — 24 packages
here plus `cuda-toolkit=12.4.*` / `cuda-version=12.4` deleted outright from
`conda.yml`.

Starting with the torch block (exactly what's on `main` today):

```suggestion
  - python=3.12.*
  - pytorch=2.5.*
  - cpuonly
  - torchaudio=2.5.*
  - torchvision=0.20.*
```

…and please put the rest back too. If the pins were genuinely blocking the
solve for `rasterio`/`geopandas`/`shapely`, that's a real problem worth its own
PR with the solver output attached — but it shouldn't land silently inside a
catchment-delineation change.

environments/environment_cpu.yml
```diff
@@ -21,40 +21,45 @@ channels:
...
+  - absl-py
+  - bokeh
+  - cachey
+  - dask
+  - gcsfs
+  - geopandas
   - gh
-  - h5netcdf=1.7.*
-  - h5py=3.15.*
+  - google-cloud-sdk
```
grey-nearing: 🔴 **Blocking — this line is why CI is red.**

```
PackagesNotFoundError: The following packages are not available from current channels:
  - google-cloud-sdk
Platform: win-64
```

([job 104938227190](https://github.com/google-research/flood-forecasting/actions/runs/35138921937/job/104938227190).)
`google-cloud-sdk` has no `win-64` conda-forge build, so the solve fails on
Windows and fail-fast cancels ubuntu and macos — meaning **the test suite never
ran on any platform for this PR**.

It also isn't needed. `gcs.py` treats the `gcloud` CLI as one of several
optional fallbacks (`shutil.which("gcloud")` guarded) and falls through to
`gcsfs`, which is already a dependency and already handles both read and write.
Dropping this line should turn CI green; if you want the CLI available for
interactive use, that belongs in developer docs, not the test environment.

Same line exists in `conda.yml` — please drop it there too.

environments/conda.yml
```diff
@@ -22,42 +22,45 @@ channels:  # Keep order
...
-  - cuda-toolkit=12.4.*
-  - cuda-version=12.4
-  - dask=2025.12.0
-  - gcsfs=2025.12.*
-  - geopandas=1.1.*
+  - python=3.12
+  - pytorch
+  - pytorch-cuda
+  - torchaudio
+  - torchvision
```
grey-nearing: 🔴 **Blocking — same unpinning problem on the GPU/dev environment.**

Beyond the pins removed here, this file also **lost `cuda-toolkit=12.4.*` and
`cuda-version=12.4` entirely** — `pytorch-cuda` is now unconstrained, so a fresh
GPU environment can pull a CUDA that doesn't match the driver on our boxes.

Please restore the pinned block from `main` (`pytorch=2.5.1`,
`pytorch-cuda=12.4`, `torchaudio=2.5.1`, `torchvision=0.20.1`,
`cuda-toolkit=12.4.*`, `cuda-version=12.4`, plus the rest), and add the four new
deps this module actually needs (`rasterio`, `shapely`, `geopandas`, `pyarrow`)
*with* pins.

Also worth a separate note in the description: `environments/environment_gpu.yml`
is added here as a **symlink** to `conda.yml`. That's probably fine, but symlinks
in a repo that gets checked out on Windows have a way of surprising people —
worth being deliberate about.

catchment_delineation/cli.py
```diff
@@ -0,0 +1,671 @@
...
+      )
+  except CatchmentCoverageError as e:
+    sys.stderr.write(f"\nCatchment Delineation Aborted: {e}\n")
+    return 1
+  finally:
+    if args.clean_cache:
+      cache_p = tiles_dir if tiles_dir else get_default_cache_dir()
+      if cache_p.exists():
+        import shutil
+        shutil.rmtree(cache_p, ignore_errors=True)
```
grey-nearing: 🔴 **Blocking — this deletes the user's own data.**

`cache_p = tiles_dir if tiles_dir else get_default_cache_dir()` means
`--tiles-dir /data/hydrosheds/tiles --clean-cache` `rmtree`s
`/data/hydrosheds/tiles`. That directory was never a cache — the user told us it
was their source data, and `--clean-cache` is documented as "purge local DEM
tile cache after completion".

Reproduced:

```
before: ['README_FROM_USER.txt', 'n40w090.npy']
$ delineate-catchment --lat 39.6828 --lon -88.7729 \
    --tiles-dir $DIR --clean-cache -o out.geojson
exit: 0 | user tiles dir still exists: False
```

Note it took the unrelated `README_FROM_USER.txt` with it — this isn't "deletes
tiles it downloaded", it's `rm -rf` on a directory we don't own.

I applied and tested this fix (suite goes to 19 passed / 1 skipped / 1 failed,
and the one failure is `test_clean_cache_flag`, which asserts the bug):

```suggestion
  finally:
    if args.clean_cache:
      if tiles_dir is not None:
        sys.stderr.write(
            "\n--clean-cache ignored: refusing to delete the user-supplied "
            f"--tiles-dir ({tiles_dir}). Only the managed download cache is "
            "ever purged.\n"
        )
      else:
        cache_p = get_default_cache_dir()
        if cache_p.exists():
          import shutil
          shutil.rmtree(cache_p, ignore_errors=True)
```

Two smaller things in the same block: this runs in a `finally`, so the cache is
also wiped when the run crashed — which is exactly when you'd want the tiles
still there for a retry. And `ignore_errors=True` means a partial delete is
silent.

catchment_delineation/cli.py
```diff
@@ -0,0 +1,671 @@
...
+      None,
+  )
+  lon_col = lon_col_arg or next(
+      (c for c in fieldnames if any(k in c.lower() for k in ("lon", "longitude", "long", "lng", "x"))),
+      None,
+  )
+  id_col = id_col_arg or next(
+      (c for c in fieldnames if any(k in c.lower() for k in ("id", "gauge_id", "station_id", "catchment_id", "hybas_id", "name"))),
+      None,
+  )
```
grey-nearing: 🟡 **Column auto-detection is loose enough to silently pick the wrong columns.**

`any(k in c.lower() for k in ("lat", "latitude", "y"))` matches *any* column
containing the letter `y`, and the longitude matcher matches any containing `x`.
First match in file order wins. Verified:

```
columns = [xmin, ymin, gauge_lat, gauge_lon]
  -> parsed coords: [(2.0, 1.0)]      # expected [(39.6, -88.7)]
```

No exception — just wrong coordinates, and therefore wrong catchments, written
to a published artifact. Bounding-box columns, projected `x`/`y`, `area_x`,
`day_of_year` are all realistic in the CSVs people will feed this.

The `id` matcher has the same problem from the other direction: `"id" in
"validated"` is `True`, so a `validated` column would be chosen as the gauge ID.

Suggested shape (not a literal patch since it's a bigger rewrite):

1. try exact matches first, in priority order: `gauge_id` / `latitude` / `lat` /
   `gauge_lat` / `CARAVAN:gauge_lat` / `y`;
2. only then fall back to substring matching, with `"y"`/`"x"` requiring a word
   boundary rather than bare containment;
3. `logger.info` the resolved column names so a wrong guess is visible in the
   log rather than only in the output geometry.

A cheap extra safety net: after parsing, assert `-90 <= lat <= 90` and
`-180 <= lon <= 180` and raise if not. That alone would have caught the
`(2.0, 1.0)` case… actually it wouldn't, which is rather the point — ordering
the exact matches first is the real fix.

catchment_delineation/cli.py
```diff
@@ -0,0 +1,671 @@
...
+  finally:
+    if args.clean_cache:
+      cache_p = tiles_dir if tiles_dir else get_default_cache_dir()
+      if cache_p.exists():
+        import shutil
+        shutil.rmtree(cache_p, ignore_errors=True)
+
+  if args.preserve_caravan_dirs or args.output_dir:
+    base_out = normalize_gcs_path(args.output_dir if args.output_dir else args.output).rstrip("/")
+    feats = result["features"] if result.get("type") == "FeatureCollection" else [result]
```
grey-nearing: 🟡 **Two rough edges in the output path, both reproduced.**

**(a) No output path at all.** `--preserve-caravan-dirs` without `-o` or
`--output-dir` gives `base_out = normalize_gcs_path(None)` → `str(None)` →
`"None"`, and the tool cheerfully writes into a literal directory called `None`
in the cwd:

```
$ delineate-catchment --csv c.csv --preserve-caravan-dirs
Saved 1 catchments to None/caravan-original/shapefiles-rederived/camels/camels_basin_shapes.geoparquet
exit: 0
```

Please make `--preserve-caravan-dirs` either require `--output-dir` (argparse
can enforce this) or default to `.`.

**(b) Everything failed.** If no gauge delineates, `feats` is empty and
`gpd.GeoDataFrame.from_features([])` a few lines down raises:

```
ValueError: Assigning CRS to a GeoDataFrame without a geometry column is not supported.
```

An unhandled traceback rather than "0 of N catchments delineated; see
<failures.csv>". Worth an early `if not feats: ... return 1`.

While you're here: the sanitising block at 512-537 is duplicated almost verbatim
at 588-612. Worth pulling into a `_sanitize_features(feats)` helper so the
schema contract lives in one place.

catchment_delineation/cli.py
```diff
@@ -0,0 +1,671 @@
...
+          completed += 1
+          if completed % max(1, total // 20) == 0 or completed == total:
+            pct = (completed / total) * 100.0
+            print(
+                f"Progress: [{completed}/{total}] catchments evaluated ({pct:.1f}%)",
+                file=sys.stderr,
+            )
+          res_feat = fut.result()
+          if res_feat is not None:
+            features.append(res_feat)
```
grey-nearing: 🟢 **Nit: output row order is non-deterministic with `--workers > 1`.**

`as_completed` yields in completion order, so two identical runs produce
GeoParquet/GeoJSON files with different row orders — which makes the published
artifacts undiffable and any checksum-based audit useless.

You already have the index: `futures = {pool.submit(...): i for i, t in
enumerate(tasks)}`. Collect `(futures[fut], res_feat)` and sort before building
the FeatureCollection, and the output becomes byte-reproducible for free.

Also: `_delineate_worker` returns `None` on failure and the result is silently
dropped, so the output has fewer rows than the input CSV with no record of
which gauges are missing. A `--failures failures.csv` (gauge_id, lat, lon,
error) would make the 26,708-basin run auditable.

catchment_delineation/delineator.py
```diff
@@ -0,0 +1,575 @@
...
+
+    # 1. Multi-Tile BFS Reverse Flow Traversal
+    q = deque([(start_key, best_r, best_c)])
+    visited_tiles: Dict[Tuple[int, int], np.ndarray] = {
+        start_key: np.zeros((TILE_CELLS, TILE_CELLS), dtype=bool)
+    }
+    visited_tiles[start_key][best_r, best_c] = True
+    total_accum = 0
+
+    while q and total_accum < max_cells:
```
grey-nearing: 🔴 **Blocking — `max_cells` truncates the watershed silently.**

When the cap is hit the loop just stops, and nothing in the returned Feature
says the basin is incomplete. Measured on a synthetic 490,700-cell basin:

```
max_cells=10M : cells=490,700  area=3330.9 km2
max_cells=100k: cells=100,000  area= 681.9 km2   <-- no warning, no flag
properties keys: ['area_km2','bbox','catchment_id','delineation_method',
                  'delineation_mode','grid_resolution','outlet',
                  'reach_attributes','tiles_spanned_count','upstream_cells_count']
```

At 37.5°N a cell is ~6,788 m², so:

| entry point | default | silently truncates above |
| --- | ---: | ---: |
| `delineate` | 50,000,000 | ~339,000 km² |
| `delineate_batch` | **5,000,000** | **~34,000 km²** |
| `delineate_dem` / `delineate_catchment` / `delineate_coordinates` | **5,000,000** | **~34,000 km²** |

`googlehydrology.delineate_dem(lat, lon)` — the call in both the README and the
Sphinx page — needs 27.3 M cells for the Rhine and 117.9 M for the Danube. Both
come back wrong, quietly.

This is inconsistent with the philosophy everywhere else in this file: you
already refuse to return a partial basin when it crosses the coverage boundary
or hits a missing tile. Hitting the cell cap is the same situation. Either

```python
if total_accum >= max_cells:
  raise CatchmentCoverageError(
      f"Watershed exceeded max_cells={max_cells:,} before terminating; "
      "the basin is larger than the traversal limit. Re-run with a higher "
      "--max-cells to get a complete delineation."
  )
```

after the loop, or at minimum set `properties["truncated"] = True` and
`logger.warning`. And please unify the 5 M / 50 M defaults — right now the same
basin gives different answers depending on which entry point you called.

catchment_delineation/delineator.py
```diff
@@ -0,0 +1,575 @@
...
+                  f"Watershed extends into missing tile {tile_key_to_filename(*nkey)} not "
+                  f"found in user-provided tiles directory ({self.tiles_dir}). "
+                  "Delineation stopped to prevent returning a partial or misleading catchment."
+              )
+            continue
+          visited_tiles[nkey] = np.zeros((TILE_CELLS, TILE_CELLS), dtype=bool)
+        else:
+          ngrid = self.get_tile(*nkey)
+          if ngrid is None:
+            continue
```
grey-nearing: 🔴 **Blocking — in the default (GCS) mode a failed tile fetch silently clips the basin.**

The `tiles_dir` branch raises `CatchmentCoverageError` — correct. The `else`
just `continue`s. Combined with `get_tile` swallowing every download exception
at `logger.debug` and memoising `None` (lines 114-118), a transient 503, an
expired token, or a rate-limit during a multi-hour Caravan run produces
plausible-looking but wrong polygons with nothing louder than a debug message.

This is exactly the failure the `tiles_dir` branch exists to prevent, and the
cloud path is the one people will actually use.

The root problem is that `get_tile` returns `None` for two very different
things: *"this tile legitimately does not exist (ocean / outside the HydroSHEDS
grid)"* — fine, `continue` — and *"the tile exists but I could not fetch it"* —
must abort. Suggest a `TileFetchError` raised by `get_tile` on fetch failure,
with `None` reserved for genuine absence.

I hit this for real. Running the benchmark with credentials that couldn't reach
GCS, every tile fetch failed, and the report said:

```
Successful Delineations : 0 / 12
Out-of-Coverage Basins  : 12 / 12 (detected & aborted cleanly)
Overall Median IoU      : 0.000
```
```
status: OUT_OF_COVERAGE: DEM tile n50w120.npy not found for coordinates (49.5956, -117.2450)...
```

Exit code 0. A total infrastructure failure reported as a clean run. A
*partial* failure would be worse: the headline IoU would be computed over a
silently biased subset.

catchment_delineation/delineator.py
```diff
@@ -0,0 +1,575 @@
...
+
+    # 2. Calculate accurate ground area summed across all visited tiles
+    total_area_km2 = 0.0
+    for (t_lat, _), mask in visited_tiles.items():
+      cell_cnt = int(mask.sum())
+      if cell_cnt > 0:
+        mid_lat = t_lat - 2.5
+        lat_scale = RES_DEG * 111.0
+        lon_scale = RES_DEG * 111.0 * math.cos(math.radians(mid_lat))
+        total_area_km2 += cell_cnt * lat_scale * lon_scale
```
grey-nearing: 🟡 **Area uses the tile's centre latitude for every cell — measured ±8 % bias.**

`mid_lat = t_lat - 2.5` applies one `cos φ` to all cells in a 5°-tall tile.
Measured on a 200-cell strip:

| position | reported km² | per-row correct | bias |
| --- | ---: | ---: | ---: |
| top of `n40` tile (lat ≈ 40.0) | 1.4000 | 1.3126 | **+6.66 %** |
| bottom of `n40` tile (lat ≈ 35.3) | 1.4000 | 1.3989 | +0.08 % |
| top of `n60` tile (lat ≈ 60.0) | 0.9000 | 0.8579 | **+4.91 %** |
| bottom of `n60` tile (lat ≈ 55.3) | 0.9000 | 0.9775 | **−7.92 %** |

Worse than the magnitude is that it's **discontinuous at tile seams**: two
basins 100 m apart on opposite sides of a 5° line get scale factors differing
by up to 15 %, so `area_km2` isn't even a continuous function of position. If
anyone maps drainage-area error globally this will show up as 5° banding.

I applied and ran this fix; residual bias drops to +0.24 % / +0.32 % on a
2,000-cell strip, and that residual is the `round(x, 1)` on the output, not the
formula:

```suggestion
    total_area_km2 = 0.0
    for (t_lat, _), mask in visited_tiles.items():
      rows = np.flatnonzero(mask.any(axis=1))
      if rows.size:
        counts = mask[rows].sum(axis=1)
        cell_lats = t_lat - (rows + 0.5) * RES_DEG
        lat_scale = RES_DEG * 111.0
        lon_scale = RES_DEG * 111.0 * np.cos(np.radians(cell_lats))
        total_area_km2 += float((counts * lat_scale * lon_scale).sum())
```

(All existing tests still pass with this applied.)

Related doc nit: the README calls this "**exact** latitude scaling" and
"**Accurate Geodesic** Area", and the Sphinx page repeats it. It's a spherical
`111 km · cos φ` approximation — about 0.5 % off WGS84 even after the above fix.
Either soften the wording or use `pyproj.Geod().geometry_area_perimeter()`.

catchment_delineation/delineator.py
```diff
@@ -0,0 +1,575 @@
...
+          logger.debug("Tile %s download skipped or failed: %s", tile_name, e)
+          if self.cache_tiles:
+            self._tile_cache[key] = None
+          return None
+
+    try:
+      arr = np.load(tile_path, mmap_mode="r")
+      if self.cache_tiles:
+        self._tile_cache[key] = arr
+      return arr
```
grey-nearing: 🟡 **No validation of what comes out of the `.npy`.**

`np.load` will happily return an `int8` array, a `float32` array, or something
that isn't 6000×6000, and the BFS will then compare `grid[nr, nc] == req_val`
against it and produce a wrong-but-plausible basin.

The concrete case: the class docstring advertises "HydroSHEDS DIR / **MERIT
flwdir**", and MERIT Hydro's `flwdir` is `int8`, where 128 wraps to −128:

```
dtype=uint8  stored=np.uint8(128)   upstream_cells=3  (expected 3)
dtype=int8   stored=np.int8(-128)   upstream_cells=2  (expected 3)  <-- NE inflow lost
```

Every NE→SW inflow silently disappears. The tiles in the bucket are fine — I
checked `n40w090.npy`: `uint8`, all eight D8 codes present — so this only bites
someone pointing `--tiles-dir` at their own MERIT-derived tiles, which the docs
actively invite.

Two lines after `np.load` would close it:

```python
if arr.dtype != np.uint8 or arr.shape != (TILE_CELLS, TILE_CELLS):
  raise ValueError(
      f"Tile {tile_path} has dtype={arr.dtype} shape={arr.shape}; "
      f"expected uint8 {(TILE_CELLS, TILE_CELLS)} D8 flow directions."
  )
```

(`.view(np.uint8)` would also work for the `int8` case specifically, but an
explicit error is friendlier than a silent reinterpretation.)

Related: `except Exception` here logs at `debug` and returns `None`, which is
how a corrupt tile becomes an invisible hole in a watershed. Worth at least
`logger.warning`.

catchment_delineation/delineator.py
```diff
@@ -0,0 +1,575 @@
...
+                f"Watershed extends north past the DEM coverage boundary ({DEM_MAX_LAT}°N) "
+                f"at longitude {t_lon + cc * RES_DEG:.4f}°. "
+                "Delineation stopped to prevent returning a partial or misleading catchment."
+            )
+          if nt_lat < -55:
+            raise CatchmentCoverageError(
+                f"Watershed extends south past the DEM coverage boundary ({DEM_MIN_LAT}°S) "
+                f"at longitude {t_lon + cc * RES_DEG:.4f}°. "
+                "Delineation stopped to prevent returning a partial or misleading catchment."
+            )
```
grey-nearing: 🟢 **Nit: hardcoded `-55` next to an interpolated `DEM_MIN_LAT`.**

`if nt_lat < -55:` uses a literal while the message it raises interpolates
`{DEM_MIN_LAT}` (−56). `is_tile_in_coverage` also hardcodes `-55`. They agree
today only by coincidence — please derive the tile bound from the constant
(`int(DEM_MIN_LAT) + 1`, or better, add an explicit `DEM_MIN_TILE_LAT_TOP = -55`
to `config.py` with a comment explaining the −56 vs −55 relationship).

Also, the two error messages report the longitude as `t_lon + cc * RES_DEG`,
which is the *current* cell's longitude in the *current* tile — for a northward
crossing that's right, but it reads as if it were the crossing point. Minor,
but the message is what someone will paste into a bug report.

catchment_delineation/delineator.py
```diff
@@ -0,0 +1,575 @@
...
+    }
+
+    return geojson_feature
+
+  def delineate_batch(
+      self,
+      coords: Iterable[Tuple[float, float]],
+      ids: Optional[Iterable[str]] = None,
+      snap_window_cells: int = 4,
+      max_cells: int = 5000000,
```
grey-nearing: 🟡 **These defaults disagree with `delineate` and with the docs.**

| parameter | here | `delineate` | CLI | docs |
| --- | ---: | ---: | ---: | ---: |
| `snap_window_cells` | 4 | 12 | 12 | 12 |
| `max_cells` | 5,000,000 | 50,000,000 | 50,000,000 | 50,000,000 |

Same mismatch in `delineate_dem`, `delineate_catchment` and
`delineate_coordinates` at the bottom of the file. So
`googlehydrology.delineate_dem(lat, lon)` and
`DemDelineator().delineate(lat, lon)` on the *same point* snap to different
cells and truncate at different sizes — a 10× difference in the size limit.

Please define the defaults once (`DEFAULT_SNAP_WINDOW_CELLS = 12`,
`DEFAULT_MAX_CELLS = …` in `config.py`) and reference them everywhere, so
there's a single place to change them and no way for them to drift apart again.

(The `max_cells` half of this is the blocking issue I flagged at line 252 — the
5,000,000 default truncates at ~34,000 km².)

catchment_delineation/tiles.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def is_tile_in_coverage(lat_top: int, lon_left: int) -> bool:
+  """Checks whether a 5x5 degree tile falls within the global DEM coverage domain.
+
+  HydroSHEDS tiles span latitudes from 56°S (-56) to 60°N (60).
+  A tile with lat_top covers [lat_top - 5, lat_top], so valid lat_top values
+  range from -50 (covers -55 to -50) down to -55 (covers -60 to -55) and up to 60 (covers 55 to 60).
+  """
+  return -55 <= lat_top <= int(DEM_MAX_LAT)
```
grey-nearing: 🟡 **Longitude is never validated — the antimeridian fails silently.**

The function takes `lon_left` and ignores it:

```
is_tile_in_coverage(0, 180)  = True   ->  n00e180.npy   # does not exist
is_tile_in_coverage(0, -185) = True   ->  n00w185.npy   # does not exist
is_tile_in_coverage(0, 360)  = True   ->  n00e360.npy   # does not exist
```

So a basin at 179.9°E that drains across 180° asks for `lon_left = 180`, the
tile load fails, and in the default GCS mode the BFS silently `continue`s
(see my comment on `delineator.py:291`) — a clipped basin with no error. Fiji,
eastern Kamchatka drainages, and a handful of Pacific islands are the realistic
cases. Same story for a coordinate accidentally supplied in 0-360 convention,
which is depressingly common in hydromet CSVs.

Whether or not you want to support true wrap-around (I'd say not, for now — the
763-tile grid doesn't wrap), the bound should be explicit so the failure is
loud. I tested this; it breaks no existing test:

```suggestion
  """Checks whether a 5x5 degree tile falls within the global DEM coverage domain.

  HydroSHEDS tiles span latitudes from 56°S (-56) to 60°N (60) and longitudes
  from 180°W (-180) to 175°E; there is no wrap-around across the antimeridian.
  A tile with lat_top covers [lat_top - 5, lat_top], so valid lat_top values
  range from -55 (covers -60 to -55) up to 60 (covers 55 to 60).
  """
  return (-55 <= lat_top <= int(DEM_MAX_LAT)
          and -180 <= lon_left <= 180 - int(TILE_DEG))
```

Worth pairing with a longitude check in `is_coord_in_coverage` (which also only
looks at latitude) so a bad `lon` is rejected at the pour point rather than
three layers down.

test/test_catchment_delineation.py
```diff
@@ -0,0 +1,505 @@
...
+    # In CI without credentials or cached tiles: generate synthetic tile immediately
+    arr = np.zeros((6000, 6000), dtype=np.uint8)
+    for r_c in [381, 1000, 2000, 3000, 4000, 5000, 5499]:
+      for c_c in [1000, 1473, 2000, 3000, 3737, 4000, 5000]:
+        arr[r_c - 1, c_c] = 4
+        arr[r_c - 2, c_c] = 4
+        arr[r_c - 1, c_c + 1] = 16
+        arr[r_c, c_c - 1] = 1
+    np.save(tile_file, arr)
+    return tile_file
```
grey-nearing: 🔴 **Blocking — this fixture writes fake DEM tiles into the real production cache.**

`target = Path(target_dir) if target_dir else cd_gcs.get_default_tiles_dir()`,
and most callers pass `target_dir=self.cache_dir`, which for a default
`DemDelineator()` **is** `~/.cache/googlehydrology/dem/` — the directory real
runs read from. After one clean `pytest` run:

```
138M  $HOME/.cache/googlehydrology/dem
n40w090.npy  n45w090.npy  n50w120.npy  n60e010.npy
```

Two consequences:

**1. It poisons production.** `get_tile` trusts any file it finds in the cache —
no provenance check, no re-validation. So after running the test suite, a real
`delineate_dem(39.6828, -88.7729)` on that machine uses a fabricated tile with
four hand-placed stream cells and returns a 0.03 km² "catchment". Silently.

**2. The suite isn't idempotent.** Run `pytest test/test_catchment_delineation.py`
twice on the same machine:

```
run 1: 20 passed, 1 skipped in 8.85s
run 2: 1 failed, 20 passed in 9.75s
  FAILED test_live_tile_delineation - assert 0.0339 > 10.0
```

`test_live_tile_delineation` skips when no real tile is cached; run 1 plants a
synthetic `n40w090.npy` there, so run 2 un-skips and asserts real-world
properties against fake data. CI is green only because GitHub runners start from
a clean `$HOME`.

The fix is to make the fixture hermetic rather than just offline — point the
cache at `tmp_path` for the whole session:

```python
@pytest.fixture(autouse=True)
def ensure_tile_fallback(monkeypatch, tmp_path):
  fake_cache = tmp_path / "dem_cache"
  fake_cache.mkdir()
  for mod in (cd_gcs, cd_cli, cd_del, cd_bm):
    monkeypatch.setattr(mod, "get_default_cache_dir", lambda: fake_cache, raising=False)
  monkeypatch.setattr(cd_config, "DEFAULT_CACHE_DIR", fake_cache)
  ...
```

(`config.DEFAULT_CACHE_DIR` is a module-level constant evaluated at import, so
it needs patching too, and `DemDelineator.__init__` reads
`get_default_cache_dir()` — both paths have to be covered.)

Separately, 36 MB × 4 of `np.zeros((6000, 6000))` per run is a lot of I/O for
unit tests. A `TILE_CELLS`-parameterised smaller tile in tests would make the
suite both faster and clearer about what it's exercising.

test/test_catchment_delineation.py
```diff
@@ -0,0 +1,505 @@
...
+  out_json = tmp_path / "test_out.geojson"
+  exit_code = main([
+      "--lat", "39.6828",
+      "--lon", "-88.7729",
+      "--tiles-dir", str(cache_dir),
+      "--clean-cache",
+      "-o", str(out_json),
+  ])
+  assert exit_code == 0
+  assert not cache_dir.exists()
```
grey-nearing: 🔴 **Blocking — this test asserts that a data-loss bug is correct behaviour.**

The test passes `--tiles-dir <cache_dir>` — i.e. it tells the tool "these are
*my* tiles, read them from here" — and then asserts
`assert not cache_dir.exists()`, locking in `rmtree` over a directory the user
owns. See my comment on `cli.py:464`; I reproduced it deleting an unrelated file
that happened to be in the directory.

The test should assert the opposite, and a second test should cover the case the
flag is actually for:

```python
def test_clean_cache_leaves_user_tiles_dir_alone(tmp_path):
  user_tiles = tmp_path / "my_tiles"
  user_tiles.mkdir()
  np.save(user_tiles / "n40w090.npy", np.zeros((6000, 6000), dtype=np.uint8))
  keep = user_tiles / "NOTES.txt"
  keep.write_text("not ours to delete")

  exit_code = main([...,"--tiles-dir", str(user_tiles), "--clean-cache", ...])
  assert exit_code == 0
  assert user_tiles.exists()
  assert keep.exists()
```

With the fix from the `cli.py` comment applied, the current test is the *only*
one in the file that fails (19 passed / 1 skipped / 1 failed) — which is a
pretty clean demonstration that it's testing the bug rather than the behaviour.

While you're in this file, a few gaps worth filling — none of them would pass
today:

* **cross-tile traversal**, the headline feature, has no test at all. I wrote
  one: two tiles, a channel over the seam, assert
  `upstream_cells_count == 11` and `tiles_spanned_count == 2`. It passes, so
  it'd be cheap to add and it'd guard the trickiest code in the module.
* `max_cells` truncation (assert it raises or sets a flag).
* `area_km2` against an analytically-known cell count.
* `is_tile_in_coverage` for out-of-range longitudes.
* tile dtype validation.

And `test_benchmark_execution` only asserts `len(df) == 2` plus two column
names — it passes with every IoU at 0. Asserting `status == "SUCCESS"` would
make it a real smoke test.

scripts/build_benchmark_dataset.py
```diff
@@ -0,0 +1,129 @@
...
+  elif area_km2 < 2500:
+    return "3_medium"
+  elif area_km2 < 10000:
+    return "4_large"
+  else:
+    return "5_macro"
+
+def main():
+  data_dir = Path("/usr/local/google/home/gsnearing/data")
+  out_file = Path("/usr/local/google/home/gsnearing/Projects/catchment-delineation/catchment_delineation/data/benchmark_basins_1000.parquet")
```
grey-nearing: 🟡 **Hardcoded personal absolute paths in the provenance script.**

```python
data_dir = Path("/usr/local/google/home/gsnearing/data")
out_file = Path("/usr/local/google/home/gsnearing/Projects/catchment-delineation/…")
```

There's no argparse at all, so this script only runs on your workstation — and
it's the script that generates the 1,200-basin benchmark the PR's headline
accuracy number depends on. Same issue in `slice_continental_dems.py`, whose
`--out-dir` default and `--tifs` fallback list are both under
`/usr/local/google/home/gsnearing/data/DEMs/`; that one produces the 763 tiles
in the bucket, so as committed nobody can reproduce or extend the dataset.

Suggest `argparse` with `--data-dir` and `--out-file` required (or defaulted to
`gs://` URIs), which also makes the provenance self-documenting.

Two substantive questions about this script while I'm here:

1. **`compute_geodesic_area` uses `111 · cos(centroid_lat)`** — the same family
   of flat-earth approximation as `delineator.py`. So the benchmark's
   "near-zero area bias" is comparing our approximation against our
   approximation. GRDC's `calculated_drain_area` is already loaded a few lines
   up; using it directly as `reference_area_km2` would make the area metric an
   independent check.
2. **`slice_continental_dems.py` defaults to `au, sa, eu, af, as` — no `na`** —
   yet the bucket has North American tiles (`n30w080` … `n45w090`) and the
   benchmark has 200 North American basins. So the published tiles weren't
   produced by the committed script. Worth reconciling before we call this the
   provenance.
3. Sampling uses `random_state=42` (good, deterministic), but **66 of the 1,200
   selected basins are at `latitude > 60°`**, outside DEM coverage by
   construction. They're scored `iou=0.0, area_bias=-100%` and pooled into the
   headline medians. In my 12-basin sample this dragged the `5_macro` tier to a
   reported median IoU of 0.493 when the one in-coverage macro basin scored
   0.986. Please filter them out here, and report accuracy over
   `status == "SUCCESS"` separately from the coverage rate in `benchmark.py`.

googlehydrology/__init__.py
```diff
@@ -12,3 +12,12 @@
...
 
+try:
+  from catchment_delineation import (
+      DemDelineator,
+      delineate_dem,
+      delineate_catchment,
+      delineate_coordinates,
+  )
+except ImportError:
+  pass
```
grey-nearing: 🟡 **This makes every `import googlehydrology` pull in shapely, and hides real failures.**

Verified:

```
$ python -c "import googlehydrology, sys; print('shapely' in sys.modules)"
True
```

`catchment_delineation/delineator.py` imports `shapely.geometry` and
`shapely.ops` at module scope, so every training run, every `run` CLI
invocation, and every unrelated unit test now pays that import cost and takes a
hard dependency on GEOS being loadable.

And `except ImportError` is the wrong net: if shapely is *installed* but its
GEOS shared library is broken (a genuinely common conda situation), that's an
`OSError`, not an `ImportError` — so it propagates and takes down
`import googlehydrology` entirely, for a module that has nothing to do with
training.

A lazy hook keeps the convenience without the coupling:

```python
def __getattr__(name):
  if name in ("DemDelineator", "delineate_dem", "delineate_catchment",
              "delineate_coordinates"):
    import catchment_delineation
    return getattr(catchment_delineation, name)
  raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
```

If you'd rather keep the eager import, at least catch `Exception` and
`logger.debug` the reason — a silent `pass` means a broken install presents as
`AttributeError: module 'googlehydrology' has no attribute 'delineate_dem'`,
which is a miserable thing to debug.

catchment_delineation/README.md
```diff
@@ -0,0 +1,279 @@
...
+---
+
+## 1. What This Module Does
+
+- **Pure DEM D8 Flow Routing**: Performs exact reverse-flow breadth-first search (BFS) graph traversal on 3 arc-second (~90m at equator, 1/1200° cell size) D8 flow direction matrices (HydroSHEDS v1.4 / MERIT Hydro).
+- **Seamless Multi-Tile Traversal**: Dynamically traverses across 5°×5° tile boundaries without edge truncation or boundary artifacts, reconstructing the true natural watershed basin geometry regardless of river size.
+- **Channel Snapping**: Automatically snaps clicked or gauge coordinates to the nearest stream channel outlet cell using a bounded local BFS connectivity scoring algorithm.
+- **Accurate Geodesic Area**: Sums raster cell ground footprints with exact latitude scaling (`lat_scale * lon_scale`) to compute geodesic drainage area in $\text{km}^2$.
+- **Fast Run-Length Vectorization**: Combines row run-length raster interval fusion with Shapely `unary_union` and adaptive boundary simplification to produce clean, valid GeoJSON Polygon / MultiPolygon geometries.
+- **Zero Heavy Dependencies**: Requires only `numpy` and `shapely` (no GDAL, GIS servers, or heavy C-extensions).
```
grey-nearing: 🟢 **Nit: "Zero Heavy Dependencies" isn't accurate.**

`_vectorize_tile_mask` imports **rasterio** (which is GDAL) as its *primary*
path — the pure-numpy run-length code is only the fallback. The CLI needs
**geopandas** (→ pyogrio/fiona → GDAL) and **pyarrow** for every output format
it advertises. And this PR adds `rasterio`, `shapely`, `geopandas` and `pyarrow`
to both conda environment files, which is the honest statement of the
dependency footprint.

Suggest something like: "Lightweight core — the delineation kernel needs only
`numpy` and `shapely`; `rasterio` is used for fast polygonisation (with a pure
numpy fallback) and `geopandas`/`pyarrow` for the CLI's GeoParquet and Shapefile
output."

The same claim appears on the Sphinx page
(`docs/source/usage/catchment_delineation.rst:18`, "no GIS servers or GDAL
runtime required") and in `CATCHMENT_DELINEATION_SUMMARY.md`.

Two other doc facts worth correcting nearby:
* §7 says the dataset is "stratified **equally** across … 5 size tiers". Actual
  counts in the committed parquet: `1_micro=45, 2_small=207, 3_medium=305,
  4_large=308, 5_macro=335`.
* "**Accurate Geodesic Area** … exact latitude scaling" — it's a spherical
  `111 km · cos φ` approximation, and currently evaluated at the tile centre
  (see my comment on `delineator.py:313`).

CATCHMENT_DELINEATION_SUMMARY.md
```diff
@@ -0,0 +1,162 @@
+# Catchment Delineation Integration Summary
+
+**Author / Agent**: Jetski  
+**Target Repository**: `google-research/flood-forecasting` (`googlehydrology`)  
+**Local Workspace**: `/usr/local/google/home/gsnearing/Projects/catchment-delineation`  
+**Git Branch**: `gsnearing-catchment-delineation`  
+**Status**: Completed, verified, and committed  
+
```
grey-nearing: 🟢 **Nit: please delete this file before merging.**

This is an agent work-log, not repo documentation. It records
`**Local Workspace**: /usr/local/google/home/gsnearing/Projects/catchment-delineation`,
`**Author / Agent**: Jetski`, and `**Git Branch**: gsnearing-catchment-delineation`
at the top level of a public Google repo, and it's already stale — §1 claims
"6 unit and integration tests" where the PR ships 21.

Everything useful in it is either in the PR description, `catchment_delineation/README.md`,
or `docs/source/usage/catchment_delineation.rst`, all of which are better places
for it. Root-level `*_SUMMARY.md` files also tend to rot quietly because nobody
has a reason to open them again.

Same category, worth a line in the description if they're intentional:
`.gitignore` gains `docs/source/tutorial/*.ipynb`, and
`environments/environment_gpu.yml` is added as a symlink to `conda.yml` —
neither is related to catchment delineation.

environments/environment_cpu.yml
grey-nearing: **Blocking: please restore the version pins in this file and in `conda.yml`.**

Every pin has been removed here. `environment_cpu.yml` is what CI installs, so
CI now resolves whatever torch/numpy/xarray happens to be newest that day, and a
green run today says nothing about tomorrow. We ship pretrained checkpoints and
retrain rarely: an unpinned `pytorch` means a checkpoint saved this week may not
load next month, and nobody will connect the failure to this PR.

The full delta vs `main`: `pytorch 2.5.*`, `torchvision 0.20.*`,
`torchaudio 2.5.*`, `numpy 2.4.*`, `pandas 2.3.*`, `xarray 2025.12.*`,
`zarr 3.1.*`, `dask`, `gcsfs`, `geopandas`, `scipy`, `sphinx`, `tensorboard`,
`bokeh`, `absl-py`, `h5py`, `h5netcdf`, `netcdf4`, `matplotlib`,
`more-itertools`, `nbsphinx`, `parso`, `pydantic`, `ruamel.yaml` — 24 packages
here plus `cuda-toolkit=12.4.*` / `cuda-version=12.4` deleted outright from
`conda.yml`.

The torch block, as on `main` today:

```suggestion
  - python=3.12.*
  - pytorch=2.5.*
  - cpuonly
  - torchaudio=2.5.*
  - torchvision=0.20.*
```

…and please restore the rest. If the pins were genuinely blocking the solve for
`rasterio`/`geopandas`/`shapely`, that deserves its own PR with the solver output
attached, not a silent change inside a catchment-delineation PR.

environments/environment_cpu.yml
grey-nearing: **Blocking: please drop this dependency, here and in `conda.yml` — it is why CI is red.**

```
PackagesNotFoundError: The following packages are not available from current channels:
  - google-cloud-sdk
Platform: win-64
```

([job 104938227190](https://github.com/google-research/flood-forecasting/actions/runs/35138921937/job/104938227190).)
`google-cloud-sdk` has no `win-64` conda-forge build, so the solve fails on
Windows and fail-fast cancels ubuntu and macos — meaning **the test suite never
ran on any platform for this PR**.

It also isn't needed: `gcs.py` treats the `gcloud` CLI as one of several optional
fallbacks (`shutil.which("gcloud")` guarded) and falls through to `gcsfs`, which
is already a dependency and already handles both read and write. Dropping this
line should turn CI green. If the CLI is wanted for interactive use, that belongs
in developer docs, not the test environment.

environments/conda.yml
grey-nearing: **Blocking: please restore the pins on the GPU/dev environment too, including the CUDA pins, and pin the four new geospatial deps.**

Beyond the pins removed here, this file also **lost `cuda-toolkit=12.4.*` and
`cuda-version=12.4` entirely** — `pytorch-cuda` is now unconstrained, so a fresh
GPU environment can pull a CUDA that doesn't match the driver on our boxes.

Please restore the pinned block from `main` (`pytorch=2.5.1`,
`pytorch-cuda=12.4`, `torchaudio=2.5.1`, `torchvision=0.20.1`,
`cuda-toolkit=12.4.*`, `cuda-version=12.4`, plus the rest), and add the four new
deps this module actually needs (`rasterio`, `shapely`, `geopandas`, `pyarrow`)
*with* pins.

Also worth a line in the description: `environments/environment_gpu.yml` is added
here as a **symlink** to `conda.yml`. That's probably fine, but symlinks in a repo
that gets checked out on Windows have a way of surprising people.

catchment_delineation/cli.py
grey-nearing: **Blocking: please scope `--clean-cache` to the managed cache — it currently deletes the user's own data.**

`cache_p = tiles_dir if tiles_dir else get_default_cache_dir()` means
`--tiles-dir /data/hydrosheds/tiles --clean-cache` `rmtree`s
`/data/hydrosheds/tiles`, which is the user's source data, not a cache;
`--clean-cache` is documented as "purge local DEM tile cache after completion".
Reproduced:

```
before: ['README_FROM_USER.txt', 'n40w090.npy']
$ delineate-catchment --lat 39.6828 --lon -88.7729 \
    --tiles-dir $DIR --clean-cache -o out.geojson
exit: 0 | user tiles dir still exists: False
```

It took the unrelated `README_FROM_USER.txt` with it: this isn't "deletes tiles it
downloaded", it's `rm -rf` on a directory we don't own.

With this applied the suite goes to 19 passed / 1 skipped / 1 failed, and the one
failure is `test_clean_cache_flag`, which asserts the bug:

```suggestion
  finally:
    if args.clean_cache:
      if tiles_dir is not None:
        sys.stderr.write(
            "\n--clean-cache ignored: refusing to delete the user-supplied "
            f"--tiles-dir ({tiles_dir}). Only the managed download cache is "
            "ever purged.\n"
        )
      else:
        cache_p = get_default_cache_dir()
        if cache_p.exists():
          import shutil
          shutil.rmtree(cache_p, ignore_errors=True)
```

Two smaller things in the same block: this runs in a `finally`, so the cache is
also wiped when the run crashed — exactly when you'd want the tiles still there
for a retry. And `ignore_errors=True` means a partial delete is silent.

catchment_delineation/cli.py
grey-nearing: **Please tighten the CSV column auto-detection — it can silently pick the wrong columns and publish wrong catchments.**

`any(k in c.lower() for k in ("lat", "latitude", "y"))` matches *any* column
containing the letter `y`, and the longitude matcher matches any containing `x`.
First match in file order wins:

```
columns = [xmin, ymin, gauge_lat, gauge_lon]
  -> parsed coords: [(2.0, 1.0)]      # expected [(39.6, -88.7)]
```

No exception — just wrong coordinates, and wrong catchments, in a published
artifact. Bounding-box columns, projected `x`/`y`, `area_x`, `day_of_year` are all
realistic in the CSVs people feed this. The `id` matcher has the same problem from
the other direction: `"id" in "validated"` is
`True`, so a `validated` column would be chosen as the gauge ID.

Suggested shape (not a literal patch since it's a bigger rewrite):

1. try exact matches first, in priority order: `gauge_id` / `latitude` / `lat` /
   `gauge_lat` / `CARAVAN:gauge_lat` / `y`;
2. only then fall back to substring matching, with `"y"`/`"x"` requiring a word
   boundary rather than bare containment;
3. `logger.info` the resolved column names so a wrong guess is visible in the log
   rather than only in the output geometry.

A post-parse range check (`-90 <= lat <= 90`, `-180 <= lon <= 180`) is a cheap
extra safety net, but note it would *not* have caught the `(2.0, 1.0)` case —
ordering the exact matches first is the real fix.

catchment_delineation/cli.py
grey-nearing: **Please fix two rough edges in the output path: no output location given, and every gauge failing.**

**(a) No output path at all.** `--preserve-caravan-dirs` without `-o` or
`--output-dir` gives `base_out = normalize_gcs_path(None)` → `str(None)` →
`"None"`, and the tool writes into a literal directory called `None` in the cwd:

```
$ delineate-catchment --csv c.csv --preserve-caravan-dirs
Saved 1 catchments to None/caravan-original/shapefiles-rederived/camels/camels_basin_shapes.geoparquet
exit: 0
```

Please make `--preserve-caravan-dirs` either require `--output-dir` (argparse can
enforce this) or default to `.`.

**(b) Everything failed.** If no gauge delineates, `feats` is empty and
`gpd.GeoDataFrame.from_features([])` a few lines down raises:

```
ValueError: Assigning CRS to a GeoDataFrame without a geometry column is not supported.
```

An unhandled traceback rather than "0 of N catchments delineated; see
<failures.csv>". Worth an early `if not feats: ... return 1`.

Separately: the sanitising block at 512-537 is duplicated almost verbatim at
588-612 — worth a `_sanitize_features(feats)` helper so the schema contract lives
in one place.

catchment_delineation/cli.py
grey-nearing: **Nit: please make output row order deterministic with `--workers > 1`, and record the gauges that failed.**

`as_completed` yields in completion order, so two identical runs produce
GeoParquet/GeoJSON files with different row orders — which makes the published
artifacts undiffable and any checksum-based audit useless. The index is already
there: `futures = {pool.submit(...): i for i, t in enumerate(tasks)}`. Collect
`(futures[fut], res_feat)` and sort before building the FeatureCollection, and the
output becomes byte-reproducible for free.

Also: `_delineate_worker` returns `None` on failure and the result is silently
dropped, so the output has fewer rows than the input CSV with no record of which
gauges are missing. A `--failures failures.csv` (gauge_id, lat, lon, error) would
make the 26,708-basin run auditable.

catchment_delineation/delineator.py
grey-nearing: **Blocking: please make hitting `max_cells` fail loudly instead of returning a silently truncated watershed, and unify the defaults.**

When the cap is hit the loop just stops, and nothing in the returned Feature says
the basin is incomplete. Measured on a synthetic 490,700-cell basin:

```
max_cells=10M : cells=490,700  area=3330.9 km2
max_cells=100k: cells=100,000  area= 681.9 km2   <-- no warning, no flag
properties keys: ['area_km2','bbox','catchment_id','delineation_method',
                  'delineation_mode','grid_resolution','outlet',
                  'reach_attributes','tiles_spanned_count','upstream_cells_count']
```

At 37.5°N a cell is ~6,788 m², so:

| entry point | default | silently truncates above |
| --- | ---: | ---: |
| `delineate` | 50,000,000 | ~339,000 km² |
| `delineate_batch` | **5,000,000** | **~34,000 km²** |
| `delineate_dem` / `delineate_catchment` / `delineate_coordinates` | **5,000,000** | **~34,000 km²** |

`googlehydrology.delineate_dem(lat, lon)` — the call in both the README and the
Sphinx page — needs 27.3 M cells for the Rhine and 117.9 M for the Danube. Both
come back wrong, quietly.

This is inconsistent with the rest of the file, which already refuses to return a
partial basin at the coverage boundary or on a missing tile. Hitting the cell cap
is the same situation. Either

```python
if total_accum >= max_cells:
  raise CatchmentCoverageError(
      f"Watershed exceeded max_cells={max_cells:,} before terminating; "
      "the basin is larger than the traversal limit. Re-run with a higher "
      "--max-cells to get a complete delineation."
  )
```

after the loop, or at minimum set `properties["truncated"] = True` and
`logger.warning`. And please unify the 5 M / 50 M defaults — right now the same
basin gives different answers depending on the entry point.

catchment_delineation/delineator.py
grey-nearing: **Blocking: please make a failed tile fetch abort in the default (GCS) mode instead of silently clipping the basin.**

The `tiles_dir` branch raises `CatchmentCoverageError` — correct. The `else` just
`continue`s. Combined with `get_tile` swallowing every download exception at
`logger.debug` and memoising `None` (lines 114-118), a transient 503, an expired
token, or a rate-limit during a multi-hour Caravan run produces plausible-looking
but wrong polygons with nothing louder than a debug message. This is exactly the
failure the `tiles_dir` branch exists to prevent, and the cloud path is the one
people will actually use.

`get_tile` returns `None` for two very different things:
*"this tile legitimately does not exist (ocean / outside the HydroSHEDS grid)"* —
fine, `continue` — and *"the tile exists but I could not fetch it"* — must abort.
Suggest a `TileFetchError` raised by `get_tile` on fetch failure, with `None`
reserved for genuine absence.

In practice: with credentials that can't reach GCS, every tile fetch fails and the
benchmark reports:

```
Successful Delineations : 0 / 12
Out-of-Coverage Basins  : 12 / 12 (detected & aborted cleanly)
Overall Median IoU      : 0.000
```
```
status: OUT_OF_COVERAGE: DEM tile n50w120.npy not found for coordinates (49.5956, -117.2450)...
```

Exit code 0 — a total infrastructure failure reported as a clean run. A *partial*
failure would be worse: the headline IoU would be computed over a silently biased
subset.

catchment_delineation/delineator.py
grey-nearing: **Please scale cell area by each cell's latitude rather than the tile's centre latitude — the bias is up to ±8 % and jumps at tile seams.**

`mid_lat = t_lat - 2.5` applies one `cos φ` to all cells in a 5°-tall tile.
Measured on a 200-cell strip:

| position | reported km² | per-row correct | bias |
| --- | ---: | ---: | ---: |
| top of `n40` tile (lat ≈ 40.0) | 1.4000 | 1.3126 | **+6.66 %** |
| bottom of `n40` tile (lat ≈ 35.3) | 1.4000 | 1.3989 | +0.08 % |
| top of `n60` tile (lat ≈ 60.0) | 0.9000 | 0.8579 | **+4.91 %** |
| bottom of `n60` tile (lat ≈ 55.3) | 0.9000 | 0.9775 | **−7.92 %** |

Worse than the magnitude is that it's **discontinuous at tile seams**: two basins
100 m apart on opposite sides of a 5° line get scale factors differing by up to
15 %, so `area_km2` isn't even a continuous function of position. If anyone maps
drainage-area error globally this will show up as 5° banding.

With this fix the residual bias drops to +0.24 % / +0.32 % on a 2,000-cell strip,
and that residual is the `round(x, 1)` on the output, not the formula:

```suggestion
    total_area_km2 = 0.0
    for (t_lat, _), mask in visited_tiles.items():
      rows = np.flatnonzero(mask.any(axis=1))
      if rows.size:
        counts = mask[rows].sum(axis=1)
        cell_lats = t_lat - (rows + 0.5) * RES_DEG
        lat_scale = RES_DEG * 111.0
        lon_scale = RES_DEG * 111.0 * np.cos(np.radians(cell_lats))
        total_area_km2 += float((counts * lat_scale * lon_scale).sum())
```

(All existing tests still pass with this applied.)

Related doc nit: the README calls this "**exact** latitude scaling" and
"**Accurate Geodesic** Area", and the Sphinx page repeats it. It's a spherical
`111 km · cos φ` approximation — about 0.5 % off WGS84 even after the above fix.
Either soften the wording or use `pyproj.Geod().geometry_area_perimeter()`.

catchment_delineation/delineator.py
grey-nearing: **Please validate the dtype and shape of the array loaded from the `.npy`, and log tile-load failures above `debug`.**

`np.load` will happily return an `int8` array, a `float32` array, or something
that isn't 6000×6000; the BFS then compares `grid[nr, nc] == req_val` against it
and produces a wrong-but-plausible basin.

The concrete case: the class docstring advertises "HydroSHEDS DIR / **MERIT
flwdir**", and MERIT Hydro's `flwdir` is `int8`, where 128 wraps to −128:

```
dtype=uint8  stored=np.uint8(128)   upstream_cells=3  (expected 3)
dtype=int8   stored=np.int8(-128)   upstream_cells=2  (expected 3)  <-- NE inflow lost
```

Every NE→SW inflow silently disappears. The tiles in the bucket are fine —
`n40w090.npy` is `uint8` with all eight D8 codes present — so this only bites
someone pointing `--tiles-dir` at their own MERIT-derived tiles, which the docs
actively invite. Two lines after `np.load` would close it:

```python
if arr.dtype != np.uint8 or arr.shape != (TILE_CELLS, TILE_CELLS):
  raise ValueError(
      f"Tile {tile_path} has dtype={arr.dtype} shape={arr.shape}; "
      f"expected uint8 {(TILE_CELLS, TILE_CELLS)} D8 flow directions."
  )
```

(`.view(np.uint8)` would also work for the `int8` case specifically, but an
explicit error is friendlier than a silent reinterpretation.)

Related: `except Exception` here logs at `debug` and returns `None`, which is how
a corrupt tile becomes an invisible hole in a watershed. Worth at least
`logger.warning`.

catchment_delineation/delineator.py
grey-nearing: **Nit: please derive this `-55` from the `DEM_MIN_LAT` constant instead of hardcoding it.**

`if nt_lat < -55:` uses a literal while the message it raises interpolates
`{DEM_MIN_LAT}` (−56). `is_tile_in_coverage` also hardcodes `-55`. They agree
today only by coincidence — derive the tile bound from the constant
(`int(DEM_MIN_LAT) + 1`, or better, add an explicit `DEM_MIN_TILE_LAT_TOP = -55`
to `config.py` with a comment explaining the −56 vs −55 relationship).

Also, the two error messages report the longitude as `t_lon + cc * RES_DEG`, the
*current* cell's longitude in the *current* tile — for a northward crossing that's
right, but it reads as if it were the crossing point. Minor, but the message is
what someone will paste into a bug report.

catchment_delineation/delineator.py
grey-nearing: **Please define these defaults once in `config.py` — they disagree with `delineate`, with the CLI, and with the docs.**

| parameter | here | `delineate` | CLI | docs |
| --- | ---: | ---: | ---: | ---: |
| `snap_window_cells` | 4 | 12 | 12 | 12 |
| `max_cells` | 5,000,000 | 50,000,000 | 50,000,000 | 50,000,000 |

Same mismatch in `delineate_dem`, `delineate_catchment` and
`delineate_coordinates` at the bottom of the file. So
`googlehydrology.delineate_dem(lat, lon)` and `DemDelineator().delineate(lat,
lon)` on the *same point* snap to different cells and truncate at different sizes
— a 10× difference in the size limit.

Define `DEFAULT_SNAP_WINDOW_CELLS = 12` and `DEFAULT_MAX_CELLS = …` in `config.py`
and reference them everywhere, so they can't drift apart again.

(The `max_cells` half of this is the blocking issue flagged at line 252 — the
5,000,000 default truncates at ~34,000 km².)

catchment_delineation/tiles.py
grey-nearing: **Please validate longitude here — it is ignored today, so the antimeridian fails silently.**

The function takes `lon_left` and ignores it:

```
is_tile_in_coverage(0, 180)  = True   ->  n00e180.npy   # does not exist
is_tile_in_coverage(0, -185) = True   ->  n00w185.npy   # does not exist
is_tile_in_coverage(0, 360)  = True   ->  n00e360.npy   # does not exist
```

So a basin at 179.9°E that drains across 180° asks for `lon_left = 180`, the tile
load fails, and in the default GCS mode the BFS silently `continue`s (see the
comment on `delineator.py:291`) — a clipped basin with no error. Fiji, eastern
Kamchatka drainages, and a handful of Pacific islands are the realistic cases.
Same story for a coordinate accidentally supplied in 0-360 convention, which is
depressingly common in hydromet CSVs.

Whether or not true wrap-around is worth supporting (probably not, for now — the
763-tile grid doesn't wrap), the bound should be explicit so the failure is loud.
This breaks no existing test:

```suggestion
  """Checks whether a 5x5 degree tile falls within the global DEM coverage domain.

  HydroSHEDS tiles span latitudes from 56°S (-56) to 60°N (60) and longitudes
  from 180°W (-180) to 175°E; there is no wrap-around across the antimeridian.
  A tile with lat_top covers [lat_top - 5, lat_top], so valid lat_top values
  range from -55 (covers -60 to -55) up to 60 (covers 55 to 60).
  """
  return (-55 <= lat_top <= int(DEM_MAX_LAT)
          and -180 <= lon_left <= 180 - int(TILE_DEG))
```

Worth pairing with a longitude check in `is_coord_in_coverage` (which also only
looks at latitude) so a bad `lon` is rejected at the pour point rather than three
layers down.

test/test_catchment_delineation.py
grey-nearing: **Blocking: please make this fixture hermetic — it writes fake DEM tiles into the real production cache.**

`target = Path(target_dir) if target_dir else cd_gcs.get_default_tiles_dir()`, and
most callers pass `target_dir=self.cache_dir`, which for a default
`DemDelineator()` **is** `~/.cache/googlehydrology/dem/` — the directory real runs
read from. After one clean `pytest` run:

```
138M  $HOME/.cache/googlehydrology/dem
n40w090.npy  n45w090.npy  n50w120.npy  n60e010.npy
```

**1. It poisons production.** `get_tile` trusts any file it finds in the cache: no
provenance check, no re-validation. After a test run, a real
`delineate_dem(39.6828, -88.7729)` on that machine uses a fabricated tile with four
hand-placed stream cells and returns a 0.03 km² "catchment". Silently.

**2. The suite isn't idempotent.** Run `pytest test/test_catchment_delineation.py`
twice on the same machine:

```
run 1: 20 passed, 1 skipped in 8.85s
run 2: 1 failed, 20 passed in 9.75s
  FAILED test_live_tile_delineation - assert 0.0339 > 10.0
```

`test_live_tile_delineation` skips when no real tile is cached; run 1 plants a
synthetic `n40w090.npy` there, so run 2 un-skips and asserts real-world properties
against fake data. CI is green only because GitHub runners start from a clean
`$HOME`.

Point the cache at `tmp_path` for the whole session:

```python
@pytest.fixture(autouse=True)
def ensure_tile_fallback(monkeypatch, tmp_path):
  fake_cache = tmp_path / "dem_cache"
  fake_cache.mkdir()
  for mod in (cd_gcs, cd_cli, cd_del, cd_bm):
    monkeypatch.setattr(mod, "get_default_cache_dir", lambda: fake_cache, raising=False)
  monkeypatch.setattr(cd_config, "DEFAULT_CACHE_DIR", fake_cache)
  ...
```

(`config.DEFAULT_CACHE_DIR` is a module-level constant evaluated at import, so it
needs patching too, and `DemDelineator.__init__` reads `get_default_cache_dir()` —
both paths have to be covered.)

Separately, 36 MB × 4 of `np.zeros((6000, 6000))` per run is a lot of I/O for unit
tests. A `TILE_CELLS`-parameterised smaller tile would make the suite both faster
and clearer about what it's exercising.

test/test_catchment_delineation.py
grey-nearing: **Blocking: please invert this test — it asserts that a data-loss bug is correct behaviour.**

The test passes `--tiles-dir <cache_dir>` — i.e. it tells the tool "these are *my*
tiles, read them from here" — and then asserts `assert not cache_dir.exists()`,
locking in `rmtree` over a directory the user owns. See the comment on
`cli.py:464`, where this is reproduced deleting an unrelated file that happened to
be in the directory.

The test should assert the opposite, and a second test should cover the case the
flag is actually for:

```python
def test_clean_cache_leaves_user_tiles_dir_alone(tmp_path):
  user_tiles = tmp_path / "my_tiles"
  user_tiles.mkdir()
  np.save(user_tiles / "n40w090.npy", np.zeros((6000, 6000), dtype=np.uint8))
  keep = user_tiles / "NOTES.txt"
  keep.write_text("not ours to delete")

  exit_code = main([...,"--tiles-dir", str(user_tiles), "--clean-cache", ...])
  assert exit_code == 0
  assert user_tiles.exists()
  assert keep.exists()
```

With the `cli.py` fix applied, this is the *only* test in the file that fails
(19 passed / 1 skipped / 1 failed).

A few gaps in this file worth filling, none covered today:

* **cross-tile traversal**, the headline feature, has no test at all. Two tiles, a
  channel over the seam, asserting `upstream_cells_count == 11` and
  `tiles_spanned_count == 2` passes as written, so it'd be cheap to add and would
  guard the trickiest code in the module.
* `max_cells` truncation (assert it raises or sets a flag).
* `area_km2` against an analytically-known cell count.
* `is_tile_in_coverage` for out-of-range longitudes.
* tile dtype validation.

And `test_benchmark_execution` only asserts `len(df) == 2` plus two column names —
it passes with every IoU at 0. Asserting `status == "SUCCESS"` would make it a real
smoke test.

scripts/build_benchmark_dataset.py
grey-nearing: **Please replace these hardcoded personal paths with arguments, and fix three problems with the benchmark this script produces.**

```python
data_dir = Path("/usr/local/google/home/gsnearing/data")
out_file = Path("/usr/local/google/home/gsnearing/Projects/catchment-delineation/…")
```

There's no argparse at all, so this script only runs on one workstation — and it
generates the 1,200-basin benchmark the PR's headline accuracy number depends on. Same issue in `slice_continental_dems.py`, whose `--out-dir`
default and `--tifs` fallback list are both under
`/usr/local/google/home/gsnearing/data/DEMs/`; that one produces the 763 tiles in
the bucket, so as committed nobody can reproduce or extend the dataset. Suggest
`argparse` with `--data-dir` and `--out-file` required (or defaulted to `gs://`
URIs), which also makes the provenance self-documenting.

1. **`compute_geodesic_area` uses `111 · cos(centroid_lat)`** — the same family of
   flat-earth approximation as `delineator.py`. So the benchmark's "near-zero area
   bias" is comparing our approximation against our approximation. GRDC's
   `calculated_drain_area` is already loaded a few lines up; using it directly as
   `reference_area_km2` would make the area metric an independent check.
2. **`slice_continental_dems.py` defaults to `au, sa, eu, af, as` — no `na`** — yet
   the bucket has North American tiles (`n30w080` … `n45w090`) and the benchmark has
   200 North American basins. So the published tiles weren't produced by the
   committed script. Worth reconciling before we call this the provenance.
3. Sampling uses `random_state=42` (good, deterministic), but **66 of the 1,200
   selected basins are at `latitude > 60°`**, outside DEM coverage by construction.
   They're scored `iou=0.0, area_bias=-100%` and pooled into the headline medians.
   In a 12-basin sample this dragged the `5_macro` tier to a reported median IoU of
   0.493 when the one in-coverage macro basin scored 0.986. Please filter them out
   here, and report accuracy over `status == "SUCCESS"` separately from the coverage
   rate in `benchmark.py`.

googlehydrology/__init__.py
grey-nearing: **Please make this import lazy — every `import googlehydrology` now pulls in shapely, and a broken install is hidden.**

Verified:

```
$ python -c "import googlehydrology, sys; print('shapely' in sys.modules)"
True
```

`catchment_delineation/delineator.py` imports `shapely.geometry` and `shapely.ops`
at module scope, so every training run, every `run` CLI invocation, and every
unrelated unit test now pays that import cost and takes a hard dependency on GEOS
being loadable.

`except ImportError` is also the wrong net: if shapely is *installed* but its GEOS
shared library is broken (a genuinely common conda situation), that's an `OSError`,
not an `ImportError` — so it propagates and takes down `import googlehydrology`
entirely, for a module that has nothing to do with training.

A lazy hook keeps the convenience without the coupling:

```python
def __getattr__(name):
  if name in ("DemDelineator", "delineate_dem", "delineate_catchment",
              "delineate_coordinates"):
    import catchment_delineation
    return getattr(catchment_delineation, name)
  raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
```

If the eager import is preferred, at least catch `Exception` and `logger.debug` the
reason — a silent `pass` means a broken install presents as `AttributeError: module
'googlehydrology' has no attribute 'delineate_dem'`, which is a miserable thing to
debug.

catchment_delineation/README.md
grey-nearing: **Nit: please reword "Zero Heavy Dependencies" — it isn't accurate — and correct two related doc claims.**

`_vectorize_tile_mask` imports **rasterio** (which is GDAL) as its *primary* path;
the pure-numpy run-length code is only the fallback. The CLI needs **geopandas**
(→ pyogrio/fiona → GDAL) and **pyarrow** for every output format it advertises. And
this PR adds `rasterio`, `shapely`, `geopandas` and `pyarrow` to both conda
environment files, which is the honest statement of the dependency footprint.

Suggest something like: "Lightweight core — the delineation kernel needs only
`numpy` and `shapely`; `rasterio` is used for fast polygonisation (with a pure numpy
fallback) and `geopandas`/`pyarrow` for the CLI's GeoParquet and Shapefile output."

The same claim appears on the Sphinx page
(`docs/source/usage/catchment_delineation.rst:18`, "no GIS servers or GDAL runtime
required") and in `CATCHMENT_DELINEATION_SUMMARY.md`.

Two other doc facts worth correcting nearby:
* §7 says the dataset is "stratified **equally** across … 5 size tiers". Actual
  counts in the committed parquet: `1_micro=45, 2_small=207, 3_medium=305,
  4_large=308, 5_macro=335`.
* "**Accurate Geodesic Area** … exact latitude scaling" — it's a spherical
  `111 km · cos φ` approximation, currently evaluated at the tile centre (see the
  comment on `delineator.py:313`).

CATCHMENT_DELINEATION_SUMMARY.md
grey-nearing: **Nit: please delete this file before merging.**

This is an agent work-log, not repo documentation. It records
`**Local Workspace**: /usr/local/google/home/gsnearing/Projects/catchment-delineation`,
`**Author / Agent**: Jetski`, and `**Git Branch**: gsnearing-catchment-delineation`
at the top level of a public Google repo, and it's already stale — §1 claims "6 unit
and integration tests" where the PR ships 21.

Everything useful in it is either in the PR description,
`catchment_delineation/README.md`, or `docs/source/usage/catchment_delineation.rst`,
all of which are better places for it. Root-level `*_SUMMARY.md` files also tend to
rot quietly because nobody has a reason to open them again.

Same category, worth a line in the description if they're intentional: `.gitignore`
gains `docs/source/tutorial/*.ipynb`, and `environments/environment_gpu.yml` is added
as a symlink to `conda.yml` — neither is related to catchment delineation.

PR #295, opened 2026-09-18
Remove 0.5 factor inside MaskedRMSELoss sqrt

### LINKED ISSUE #276: [BUG] MaskedRMSELoss includes 0.5 factor inside sqrt, causing numerical divergence from standard RMSE metrics
**Location**
`googlehydrology/training/loss.py:306-311`, `test/test_losses.py:127`

**Root Cause Mechanism**
`MaskedRMSELoss._get_loss` applies a $0.5$ factor inside the radical:

$$\text{loss} = \sqrt{0.5 \times \text{mean}\left((\hat{y} - y)^2\right)} = \frac{1}{\sqrt{2}}\,\text{RMSE} \approx 0.7071 \times \text{RMSE}$$

While the $\frac{1}{2}$ multiplier is standard in Mean Squared Error objectives ($\frac{1}{2}\text{MSE}$) to cancel the exponent during differentiation, placing it inside the square root scales the loss value down by $\approx 29.3\%$. Consequently, it mismatches `googlehydrology.evaluation.metrics.rmse`, which computes standard $\sqrt{\text{MSE}}$ without the scaling factor.

**Impact**
Training with `loss: 'rmse'` optimizes an artificially attenuated loss landscape, producing loss curves and training checkpoints whose numerical magnitudes diverge from all standard validation and benchmark evaluation metrics.

**Suggested Fix**
Remove the `0.5 *` coefficient in `MaskedRMSELoss._get_loss` and align the expected baseline in `test/test_losses.py`:

```python
loss = torch.sqrt(
    torch.mean((prediction['y_hat'][mask] - ground_truth['y'][mask]) ** 2)
)

```

**Verification Checklist**

* [ ] Update test assertions in `test/test_losses.py:127` to reflect standard unscaled RMSE values.
* [ ] Confirm that `MaskedRMSELoss` output exactly matches `googlehydrology.evaluation.metrics.rmse` on identical masked tensors.
* [ ] Verify learning rate dynamics or optimizer schedulers configured for RMSE loss runs to account for the $\approx 1.414\times$ change in effective loss and gradient magnitude.

dafahaha:
Fixes #276

The 0.5 factor was inside the sqrt, making MaskedRMSELoss output ~0.707x of standard RMSE. This aligns the training loss with the evaluation metric.


sylvesterkaczmarek (commented):
loss.py doesn't parse on this head: the new mask = ...        y = ... line puts two statements together with no separator. This needs to be split back into two lines before merge.

grey-nearing:
Thanks for the contribution! Note that this diff removes `0.5` from `MaskedMSELoss` rather than `MaskedRMSELoss`, and introduces syntax errors on lines 310 (unclosed parenthesis in `torch.sqrt`) and 399 (two statements merged onto one line in `MaskedCMALLoss`). More broadly, we are keeping the current `MaskedRMSELoss` definition (`sqrt(MaskedMSELoss)`) to maintain backward compatibility with existing configurations and upstream NeuralHydrology, so we'll go ahead and close this PR.

PR #299, opened 2026-09-19
Log epoch-level gradient clipping statistics

### LINKED ISSUE #298: Log how often gradient clipping engages
**Please log how often gradient clipping actually engages, so the threshold stops being invisible.**

Right now `clip_gradient_norm` is applied silently, and there is no way to tell from a training run whether it fired on every step, on none, or somewhere in between. That makes it impossible to know whether the setting is acting as a safety net or as a step-size limiter.

### Why this came up

While reviewing #290 we measured gradient norms on a freshly-initialised `nn.LSTM(64)` + CMAL head (bs=256, T=8, synthetic inputs):

| | pre-clip gradient norm | `clip_gradient_norm: 1` |
|---|---|---|
| Loss as on `main` | **1.223** | clipping **engages** |
| Loss as in #290 | **0.153** | clipping **does not engage** |

Two things stand out, and neither was noticed by anyone:

1. Under the current loss, clipping was **binding during normal training at initialisation** — so it has been acting as a step-size limiter rather than an explosion guard, whatever the intent.
2. #290 changes loss normalisation to per-observation, which shifts gradient norms by roughly `1/predict_last_n`. After it merges, clipping will quietly stop engaging for every existing CMAL recipe.

Neither of those is visible in any log. That's the actual problem — more than the choice of threshold.

### Suggested implementation

[`googlehydrology/training/basetrainer.py:484-488`](https://github.com/google-research/flood-forecasting/blob/main/googlehydrology/training/basetrainer.py#L484-L488):

```python
if self.cfg.clip_gradient_norm is not None:
    self.scaler.unscale_(self.optimizer)
    torch.nn.utils.clip_grad_norm_(
        self.model.parameters(), self.cfg.clip_gradient_norm
    )
```

`torch.nn.utils.clip_grad_norm_` already **returns the total norm before clipping**; the return value is currently discarded. So this is close to free:

- Capture the returned norm.
- Count steps where it exceeds `clip_gradient_norm`.
- At epoch end, log something like `gradient clipped 1832/2048 steps (89%), median norm 1.41`.

Logging a couple of percentiles of the norm distribution alongside the count would also make it possible to choose the threshold from data rather than inheritance.

### Why it's worth doing

It makes the setting self-diagnosing. If it never fires, it's a safety net and is doing its job quietly. If it fires on most steps, it's a regularizer and should be recognised as one. Either is a legitimate choice — silently not knowing which one you have is not.

It would also have caught the #290 scale change automatically, rather than by manual measurement during review.

### Scope

Not a blocker for #290, and deliberately not being asked of that PR's author — this is trainer-level instrumentation, separate from the loss fix.

<sub>Raised from the #290 review. Analysis assisted by AI, reviewed by me.</sub>

Junyi-Zheng:
Fixes #298

Training currently discards the norm returned by clip_grad_norm_, leaving gradient clipping frequency unreported. This change records the pre-clip norm on every clipping call and logs epoch-level clipping counts, fractions, and norm percentiles (median, P90, P99), both in training logs and TensorBoard.

Non-finite norms are counted separately and excluded from the fraction and percentiles; epochs without finite norms explicitly report unavailable statistics. Collection is independent of loss-log sampling and transfers the detached scalars to CPU once per epoch. The optimizer, loss, clipping threshold and existing loss-log counters are unchanged.

Validation includes controlled-gradient tests with and without GradScaler, exact optimizer-update comparisons, NaN/Inf and empty-epoch cases, threshold equality and double precision, missing gradients, real TensorBoard events, CMAL training/evaluation, checkpoint continuation and frozen-module finetuning.

- Full CPU suite with coverage: **244 passed, 1 deselected** on both macOS ARM64 and Linux ARM64 (Python 3.12, PyTorch 2.5.1). Linux h5py/netCDF4 use a shared system HDF5 library, avoiding a conflict between the pip wheels.
- All 24 added executable lines and all 6 branches of the new logging method are covered. Five deliberately introduced statistical defects are rejected by the regression tests.
- Wheel/sdist build and installed-wheel smoke check pass. Sphinx introduces no new warnings relative to main; its existing missing `_static` warning remains.


grey-nearing (approved):
Thanks for fixing #298! Clean implementation and thorough test coverage. Added a tiny commit to condense the config.rst bullet and cast percentile outputs to Python floats.

